PR #48, opened 2026-05-22
Redesign Models settings with simple/advanced modes and live model chains

MONKE2525E:
# Summary

Replaces the confusing nested-accordion Models settings UI with a clean two-mode design. Simple mode (default) shows a flat model picker with Accuracy/Efficiency badges and subtle group separators — approachable for new users. Advanced mode reveals the full fallback chain editor, custom model IDs, and per-provider controls, toggled via a new switch in General settings.

Also wires up the backend model chain so the pipeline now resolves `transcription_fallback_models` / `cleanup_fallback_models` at runtime and passes the actual model string to each API call, replacing previously hardcoded model names.

## Type of Change

- [ ] Bug fix
- [x] Feature
- [x] UI change
- [ ] Refactor
- [ ] Documentation
- [ ] Test-only change
- [ ] Build or release change

## Testing

- [x] `npm run check`
- [ ] `npm run lint`
- [x] `npm run test:rust`
- [ ] `npm run test:smoke`
- [ ] `npm run test:smoke:state`
- [x] Playwright or manual UI verification, if applicable

Playwright automation verified: tile open/close toggle, Add Fallback flow, custom model row appearance, fallback chain chip rendering, summary item persistence, and simple-mode model selection. All 75 Rust tests pass.

## Privacy and Security

- [x] No API keys, personal data, local private paths, screenshots with private text, or sensitive logs are included.
- [x] API keys remain outside SQLite and are not logged.
- [x] Clipboard, hotkey, database, or provider behavior changes are explained below.

Notes: The pipeline now iterates the user-configured model chain (`transcription_fallback_models`, `cleanup_fallback_models`) at runtime. Each entry is validated by `parse_model_id` before use. No new network surface — same providers, same endpoints.

## UI Evidence

Screenshots taken during Playwright verification showing collapsed tile summary chips, open simple-mode picker with dot selector and badges, and open advanced-mode provider grid with fallback chain.

## Reviewer Notes

- `validate_setting` in `commands/mod.rs` now accepts `advanced_model_ui` as a boolean key — previously any unrecognised key was silently rejected, which was why the toggle appeared broken.
- Gateway provider has been removed from the frontend; it had no backend support and was storing models in OpenAI's slot.
- Smoke test `playwright-test-state.cjs` updated: removed the redundant second `.tile-head` click that assumed non-toggle behaviour; the tile-head is now a proper toggle button.
- Google preset models updated to `gemini-3.5-flash` (accuracy) / `gemini-2.5-flash` (efficiency).

PR #50, opened 2026-05-23
feat: models picker UX redesign — fallbacks, tier pills, animations

MONKE2525E:
## Thinking Path

> - Subsystem: **settings UI** — specifically the Models section where users choose transcription and cleanup models per provider
> - The previous models redesign (PR #48) introduced a simple/advanced mode split, but the simple mode silently cleared all fallbacks on every click, making fallback selection inaccessible unless the user found the "Advanced model settings" toggle buried in General
> - For 0.11.0 the models picker needs to be both discoverable and capable — fallbacks must be first-class, not hidden behind a toggle
> - This PR rebuilds the models picker UX in a single unified view: flat list always visible, fallbacks always selectable, advanced toggle repurposed to only show custom model inputs
> - The benefit is a picker that matches the visual language of the rest of the settings (General-tab separator lines, rounded inner container) and gives users clear affordances for quality tiers and fallback ordering

## What Changed

- **Fallback selection restored** — `toggleModelSelection` cycle logic now runs in all modes: `none` → active (or fallback if active exists) → `fallback` → remove; clicking an active model with no fallbacks now deselects it (sets default to `''`) instead of being a no-op
- **Advanced toggle relocated** — removed from `GeneralSection.svelte`, added as a footer row in `ModelsSection.svelte`; now only gates custom model name inputs, not the entire UI mode
- **Custom model delete** — hover the dot on any custom model row to reveal a CSS X button; `removeCustomModel()` promotes fallbacks or falls back to the provider's standard recommended model before pruning `models_by_provider`
- **Tier pills** — unselected recommended models show `Accurate` / `Efficient` labels; hovering the **entire row** cross-fades the label to `Add fallback` using the CSS grid stacking trick (both spans share `grid-area: 1 / 1`)
- **Pill enter/exit animation** — custom `pillScale` transition animates `max-width` from `0` to the element's natural width + opacity, giving a horizontal grow/shrink without the `scaleX` flip artefact
- **Row selection glow** — background transition slowed to `380 ms`; dot `box-shadow` blooms at `420 ms` so the glow ring visibly expands after the fill colour arrives
- **Custom model animations** — individual custom rows use `fly y:5` in/out; the custom input block uses `transition:slide` (240 ms) when the toggle fires
- **Outer tile layout** — matches General tab: flat `border-top` rows, no card box border, chevron given `10 px` right padding to clear the panel edge; `advanced-toggle-row` gets `border-top` + `border-bottom` as the last item
- **Inner model container** — `border-radius: 10 px` rounded container with gradient pseudo-element dividers between provider sections; `6 px` horizontal padding on `simple-group` so active/fallback row highlights never bleed to the container edge
- **`migrateAndLoad` fix** — `advanced_model_ui` now loaded via a dedicated `get_setting` call alongside `get_all_settings`, so it resolves correctly in mock environments where `get_all_settings` does not return that key

## Verification

- Open Settings → Models: Transcription and Clean-up tiles with separator lines matching General tab
- Open a tile: rounded container, Groq / OpenAI / Google sections with gradient dividers
- Click an unselected model → it becomes active (orange dot, `Active` pill animates in)
- Click a second model → it becomes F1 (dimmer dot, `F1` pill animates in widthwise)
- Click F1 model → fallback removed; click active model → deselects (all dots go hollow)
- Hover any unselected row → `Accurate`/`Efficient` cross-fades to `Add fallback`
- Enable Custom models toggle → inputs slide in per provider; add a model → row flies in; hover its dot → X appears; click → row flies out
- `npm run check` passes (0 errors)
- Smoke test: `node tests/smoke/playwright-test-state.cjs` (requires Tauri at port 1420)

## Risks

- `toggleModelSelection` now allows an empty active model (`''`) — the pipeline backend falls back to its hardcoded defaults when no valid model ID is stored, which is safe but means a user who deselects everything will get the Groq default silently
- Removing the `advanced_model_ui` row from General changes the General section's positional invoke indices (8 → removed, 9 → 8, 10 → 9); the indices in `loadSettings` were updated accordingly — regression risk is low but worth eyeballing in General
- Smoke test `playwright-test-state.cjs` relies on `advanced_model_ui: true` from the mock; since we now load this via `get_setting` (which the mock handles correctly), the test should continue to pass

## Model Used

- Anthropic Claude Sonnet 4.6 (claude-sonnet-4-6) via Claude Code — tool use, multi-file edit, Playwright browser automation for visual verification

### DIFF
--- a/src/lib/components/settings/GeneralSection.svelte
+++ b/src/lib/components/settings/GeneralSection.svelte
@@ -25,7 +25,6 @@
   let cleanup = $state(true);
   let contextualCaps = $state(true);
   let autoSpacing = $state(true);
-  let advancedModelUi = $state(false);
   let hotkey = $state(['ControlLeft', 'MetaLeft']);
   let recordingHotkey = $state(false);
   let capturedKeys = $state<string[]>([]);
@@ -95,7 +94,6 @@
       invoke<boolean | null>('get_setting', { key: 'cleanup_enabled' }),
       invoke<boolean | null>('get_setting', { key: 'contextual_caps_enabled' }),
       invoke<boolean | null>('get_setting', { key: 'auto_spacing_enabled' }),
-      invoke<boolean | null>('get_setting', { key: 'advanced_model_ui' }),
       invoke<string[]>('get_microphones'),
       invoke<string | null>('get_setting', { key: 'microphone_device' }),
     ]);
@@ -108,7 +106,6 @@
     cleanup = val<boolean | null>(5, null) ?? true;
     contextualCaps = val<boolean | null>(6, null) ?? true;
     autoSpacing = val<boolean | null>(7, null) ?? true;
-    advancedModelUi = val<boolean | null>(8, null) ?? false;
 
     const hk = val<string[] | null>(2, null);
     if (hk && hk.length === 2) hotkey = hk;
@@ -123,8 +120,8 @@
       selectedLanguage = language;
     }
 
-    microphones = val<string[]>(9, []);
-    selectedMic = val<string | null>(10, null) ?? '';
+    microphones = val<string[]>(8, []);
+    selectedMic = val<string | null>(9, null) ?? '';
 
     results.forEach((r, i) => {
       if (r.status === 'rejected') console.error(`GeneralSection: invoke[${i}] failed:`, r.reason);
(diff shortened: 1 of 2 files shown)


MONKE2525E:

MONKE2525E:

MONKE2525E:

MONKE2525E:

PR #58, opened 2026-05-23
feat(security): encrypt API keys in Windows Credential Manager

MONKE2525E:
## Thinking Path

> - The data layer stores all settings (providers, hotkeys, tone profiles, and API keys) in `tauri-plugin-store` — a JSON file in `%APPDATA%`
> - API keys (Groq, OpenAI, Google) were written as plaintext strings into `settings.json`, readable by any process with access to the user's AppData directory
> - API keys should be treated as secrets, not config — plaintext file storage is inappropriate for credentials
> - This PR migrates key storage to the Windows Credential Manager (DPAPI-backed per-user encryption) via the `keyring` crate with the `windows-native` backend
> - A one-shot migration runs on the first launch after upgrade: reads any plaintext keys from `settings.json`, writes them to Credential Manager, and scrubs the plaintext copies
> - The benefit is that API keys are now encrypted at rest and visible only to the authenticated Windows user, with zero changes to the frontend IPC surface

## What Changed

- Added `keyring = { version = "3", features = ["windows-native"] }` dependency — uses the Windows Credential Manager (wincred) backend backed by DPAPI
- New `src-tauri/src/data/credentials.rs` module with `get`, `set`, `has`, and `migrate_from_store` — single owner of all Credential Manager I/O
- `save_api_key` command now routes through `credentials::set()` instead of `store.set()` — keys never touch `settings.json`
- `get_api_key_status` command checks `credentials::has()` instead of `store.get().is_some()`
- `load_pipeline_config` reads keys via `credentials::get()` instead of the store — pipeline behaviour unchanged
- `migrate_from_store` called once in app setup: moves plaintext keys from `settings.json` to Credential Manager, sets `credentials_migrated_v1` flag, and saves — skipped on all subsequent launches
- Added `CREDENTIALS_MIGRATED` store constant for the one-shot flag

## Verification

- Save a key through Settings › API Keys → open **Control Panel › Credential Manager › Windows Credentials** and confirm a Generic credential with target `open-flow` appears
- Confirm `%APPDATA%\com.openflow.app\settings.json` contains no `api_key_*` fields after saving
- Trigger a dictation — transcription and cleanup should succeed (keys are resolved from Credential Manager in `load_pipeline_config`)
- **Migration path:** manually add `"api_key_groq": "sk-test"` to `settings.json`, relaunch — key migrates to Credential Manager, plaintext copy is removed, `credentials_migrated_v1: true` is written; relaunch again — migration is skipped (flag present)
- `npm run check` and `npm run lint` pass

## Risks

- **Migration safety:** if a keyring write fails during migration the plaintext copy is left intact and the done-flag is not set, so it retries next launch — no key loss on failure
- **Mock backend (addressed):** keyring v3 without `features = ["windows-native"]` silently uses a file-based mock store; this was caught in testing and the feature flag is now explicit
- **No UI surface change:** frontend still calls `save_api_key` / `get_api_key_status` with identical signatures — zero frontend changes required
- **Windows-only scope:** Open Flow targets Windows exclusively; no cross-platform credential fallback is needed

## Model Used

- Anthropic Claude Sonnet 4.6 (claude-sonnet-4-6) via Claude Code CLI — tool use, extended reasoning

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2245,18 +2245,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "keyring"
-version = "3.6.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eebcc3aff044e5944a8fbaf69eb277d11986064cba30c468730e8b9909fb551c"
-dependencies = [
- "byteorder",
- "log",
- "windows-sys 0.60.2",
- "zeroize",
-]
-
-[[package]]
 name = "leb128fmt"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2890,7 +2878,6 @@ dependencies = [
  "chrono",
  "cpal",
  "hound",
- "keyring",
  "log",
  "nnnoiseless",
  "reqwest 0.12.28",
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2245,6 +2245,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "keyring"
+version = "3.6.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eebcc3aff044e5944a8fbaf69eb277d11986064cba30c468730e8b9909fb551c"
+dependencies = [
+ "byteorder",
+ "log",
+ "windows-sys 0.60.2",
+ "zeroize",
+]
+
+[[package]]
 name = "leb128fmt"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2878,6 +2890,7 @@ dependencies = [
  "chrono",
  "cpal",
  "hound",
+ "keyring",
  "log",
  "nnnoiseless",
  "reqwest 0.12.28",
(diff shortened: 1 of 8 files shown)


MONKE2525E:

MONKE2525E:

MONKE2525E:

MONKE2525E:
@gemini-code-assist Applied the following from your last review:

- **Efficient `has()`**: reimplemented to call `CredReadW` directly and free the credential without decoding the blob — avoids a full UTF-16 decode on every existence check.
- **Store constants**: replaced all hardcoded provider strings (`"groq"`, `"openai"`, `"google"`) with `store::GROQ/OPENAI/GOOGLE` and key names with `store::KEY_GROQ` etc. throughout `credentials.rs`, `store.rs`, and `commands/mod.rs`. Provider validation in `save_setting` now uses `store::PROVIDERS.contains()`.

Intentionally skipped:
- **Caching keys from `load_pipeline_config`** — this is called once per pipeline run, not in a tight loop. Caching would add complexity without a meaningful perf benefit given the actual call frequency.
- **Migration error-handling comment** (re: `has` returning false on system error) — the migration already handles this correctly: a system read error will cause `get()` to return empty, but the error is logged and the plaintext key is only deleted after a successful write, so no data is lost.
- **`set` empty key / `unwrap` feedback** — referred to the old `keyring` code that was already replaced.

PR #59, opened 2026-05-23
feat(cache): 200-char limit, 2-day inactivity prune, and snippet exemption

MONKE2525E:
## Thinking Path

> - The `cleanup_cache` table stores LLM-processed results keyed on normalized raw transcription text, so identical inputs skip the API round-trip
> - Two silent bugs existed: long inputs (>200 chars) were cached even though verbatim repetition of a long phrase is essentially never seen again, wasting space; and items unused for 2+ days were never deleted because the prune query only checked `expires_at`, making `last_hit_at` a no-op
> - Snippets are user-defined triggers meant to be reused repeatedly, so they must be exempt from both the length gate and the inactivity prune
> - This PR adds an `is_snippet` boolean column (migration v5), gates inserts at 200 chars (with snippet bypass), fixes the prune query to include a `last_hit_at`-based condition, and wires pruning into every dictation

## What Changed

- Added `is_snippet INTEGER NOT NULL DEFAULT 0` column to `cleanup_cache` schema and migration v5 in `db.rs`
- Updated `CleanupCacheEntry` struct and `cleanup_cache_get_active` SELECT to include `is_snippet`
- Added `is_snippet: bool` parameter to `cleanup_cache_insert_new`; all call sites (including 8 test sites) updated
- Fixed `cleanup_cache_prune_expired` — now also deletes rows where `last_hit_at <= now - 2 days AND is_snippet = 0`
- Added 200-char gate in `pipeline.rs`: `allow_cache = ... && (raw.chars().count() <= 200 || has_snippets)`
- `has_snippets` (`!snippet_instructions.is_empty()`) passed as `is_snippet` flag on insert so snippet-derived entries are always retained
- Prune now runs at the start of every `run_cleanup_and_snippets` call, not only at app startup

## Verification

- `npm run test:rust` — 75 passed, 0 failed
- `npm run lint` — 0 errors, 0 warnings (svelte-check + cargo clippy)
- Manual: dictate a short phrase (<200 chars, no snippet) → confirm row in `cleanup_cache` with `is_snippet = 0`
- Manual: dictate a phrase >200 chars with no snippet trigger → confirm row is **not** inserted
- Manual: add a snippet trigger, dictate a long phrase containing it → confirm row inserted with `is_snippet = 1`
- Manual: set `last_hit_at` to 3 days ago via SQLite CLI, then dictate anything → confirm stale row is deleted

## Risks

- **Migration safety:** uses existing `ensure_table_column` pattern with `BEGIN/COMMIT/ROLLBACK` — consistent with migrations v1–v4; safe to run on existing DBs
- **Existing cache entries:** will get `is_snippet = 0` by default from the migration; they remain eligible for 2-day inactivity pruning, which is the correct behavior
- **No breaking changes** to the public `CleanupCacheEntry` struct consumers — `is_snippet` is additive

## Model Used

- Anthropic Claude Sonnet 4.6 (`claude-sonnet-4-6`) via Claude Code CLI, tool use + extended context

## Checklist

- [x] Risks documented above

### DIFF
--- a/src-tauri/src/data/db.rs
+++ b/src-tauri/src/data/db.rs
@@ -125,7 +125,7 @@ pub fn open(path: &str) -> Result<Db> {
         // explicit ROLLBACK. If the migration fails mid-way, sqlite3_exec aborts but
         // leaves the BEGIN open. Without the ROLLBACK cleanup, every subsequent INSERT
         // would execute inside that ghost transaction and be silently discarded on
-        // connection close â€” causing all user data to vanish on restart.
+        // connection close — causing all user data to vanish on restart.
         let _ = conn.execute_batch(
             "ALTER TABLE snippets ADD COLUMN instructions TEXT NOT NULL DEFAULT '';",
         );
--- a/src-tauri/src/data/db.rs
+++ b/src-tauri/src/data/db.rs
@@ -1,4 +1,4 @@
-﻿use anyhow::Result;
+use anyhow::Result;
 use rusqlite::{params, Connection};
 use serde::{Deserialize, Serialize};
 use std::sync::{Arc, Mutex, MutexGuard};
--- a/src-tauri/src/data/db.rs
+++ b/src-tauri/src/data/db.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+﻿use anyhow::Result;
 use rusqlite::{params, Connection};
 use serde::{Deserialize, Serialize};
 use std::sync::{Arc, Mutex, MutexGuard};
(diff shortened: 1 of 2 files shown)


MONKE2525E:
@gemini-code-assist Gemini's two db.rs suggestions (BOM removal and corrupted em-dash fix) were both correct and already applied via GitHub's suggestion workflow — thanks for those.

For the prune latency issue: removed the `cleanup_cache_prune_expired` call from the dictation path entirely. It was already running at app startup in `main.rs` right after `db::open`, so the per-dictation call was redundant. The startup call covers the inactivity pruning guarantee without touching the hot path.

PR #83, opened 2026-06-02
Fix contextual capitalization and punctuation handling

MONKE2525E:
﻿## What changed
- Reworked contextual capitalization so blank and unknown caret contexts capitalize instead of inheriting stale lowercase history.
- Replaced the old tail-only spacing path with a probe-driven context read and a punctuation-prefix classifier, so periods, commas, quotes, dashes, and ellipses behave correctly at injection time.
- Fixed the hotkey history reset to keep backspace tracking aligned with the current `injection.rs` API.

## Why
- The old injection path was using stale or overly broad context signals, which made empty fields and punctuation-heavy dictation unreliable.
- The new probe-first logic keeps capitalization and spacing decisions tied to the actual caret-local state instead of past injection history.

## Tests
- `cargo test --manifest-path src-tauri/Cargo.toml`


MONKE2525E:

MONKE2525E:

MONKE2525E:

MONKE2525E:
@gemini-code-assist please re-review. Latest commit adds macOS context-aware injection: AX probe via `context_probe_macos.rs`, clipboard-sniff fallback (`ClipboardSniff` source) for Electron/AX-blind apps, and CGEventTap history tracking. Previous feedback has been addressed: LAST_INJECTION fallback is wired for both platforms, enums derive Copy, double-spacing guard is in place, and windows::core::BOOL is used throughout.

MONKE2525E:
@gemini-code-assist please re-review. All 5 issues from your last review have been addressed:

1. **Newline detection bug** (HIGH): `classify_context_tail` now checks `\n`/`\r` before `is_invisible_prefix_char` so newlines are correctly treated as sentence boundaries.
2. **macOS newline filtering**: `mac.rs` history filter now passes `\n`/`\r` through for parity with Windows.
3. **Clipboard pollution**: macOS `inject_text` now clears the clipboard when the original was empty/non-text.
4. **Soft-punctuation capitalization**: `SoftPunctuationPrefix` and `InvisibleOrAmbiguousPrefix` at sentence boundaries now actually capitalize the first cased character (was logging capitalized but not transforming).
5. **Proper noun regression**: `is_safe_lowercase_candidate` now uses an explicit safe-word list so Title-Case proper nouns (London, Monday, Google) are never lowercased in continuation context.

The LAST_INJECTION fallback (mentioned as high priority) is already implemented via `fallback_probe_from_history` — the suggestion in your review referenced the old `auto_learn::InjectionProbeSource` API which was refactored into `context_probe::ContextProbeSource`.

MONKE2525E:
@gemini-code-assist please re-review. Fixed the buffer overread in `c_buf_to_string` (commit `3146bd3`): replaced `CStr::from_ptr` with an explicit null-byte scan bounded by the slice length, so the buffer is never read past its end even if the ObjC shim omits a null terminator. The previous 5 fixes (newlines, clipboard, casing, proper nouns) are in commit `95b092b`.

MONKE2525E:
@gemini-code-assist please re-review. All 3 issues from your latest review have been addressed in commit `ac2770c`:

1. **`of_write_cf_string` security (HIGH)**: Now checks the return value of `CFStringGetCString` and explicitly sets `dest[0] = '\0'` on failure, preventing a buffer overread when conversion fails mid-write.
2. **Lookbehind range calculation**: Replaced the duplicated ternary expressions with a named `CFIndex start` variable — cleaner and avoids evaluating the condition twice.
3. **Child PID count clamp**: `count` in `memory.rs` is now `bytes.min(buf.len())` to make the safety boundary explicit.

MONKE2525E:
@gemini-code-assist Please review the latest fixes. We have addressed the two outstanding issues: 1) removed the 48.0 dock fallback height when the Dock is hidden or placed on a side (height of 0.0), and 2) hidden the clipboard-sniffing fallback behind a settings key `macos_clipboard_sniff_enabled` (disabled by default) and added a toggle inside the General settings section.

MONKE2525E:
@gemini-code-assist please re-review. We have addressed the latest review feedback by directly mutating Svelte 5 reactive properties and adding a check to skip macOS clipboard sniffing if the clipboard contains non-text formats to prevent data loss.

MONKE2525E:
@gemini-code-assist please re-review. We have addressed the latest feedback by: 1) replacing the is_multiple_of check with standard modulo under a Clippy override in mac.rs, and 2) renaming the bytes variable to num_pids in memory.rs to match the actual semantics of the libproc call.

MONKE2525E:
@gemini-code-assist please re-review. We have addressed the latest feedback by: 1) refining pasteboard_has_non_text_formats to check specifically for rich or binary formatting (protecting HTML, RTF, images, PDFs, files, webarchives, while allowing standard plain text copies), and 2) correcting apply_probe_adjustments to always format and capitalize HardSentenceTerminator prefixes.

MONKE2525E:
@gemini-code-assist please re-review

MONKE2525E:
@gemini-code-assist please re-review

MONKE2525E:
@gemini-code-assist please re-review

MONKE2525E:
@gemini-code-assist please re-review

MONKE2525E:
@gemini-code-assist please re-review

PR #90, opened 2026-06-05
fix(snippets): fix paste-to-create on macOS and harden modal interactive controls

MONKE2525E:
## Thinking Path

> - Open Flow is a macOS AI dictation app — hotkey hold → audio capture → transcription → LLM cleanup → text injection
> - The Snippets and Dictionary views both present a modal dialog for creating/editing records, with auto-growing textareas
> - On macOS WKWebView, creating a snippet by **pasting** text (rather than typing) silently did nothing: the modal closed, nothing was saved, and no error appeared
> - Root cause: `backdrop-filter: blur()` on the full-screen backdrop button repaints on every layout change (e.g. textarea auto-grow); when it repaints, WKWebView's pointer hit-testing breaks and clicks fall through the card to the backdrop, which fires `closeModal()` — completely bypassing `saveModal()`
> - A second trigger existed at larger content sizes: once the modal card grew tall enough to scroll, the `position: fixed + translate(-50%, -50%)` centering broke hit-testing again even without blur
> - This PR removes `backdrop-filter` from both modal backdrops, rewrites the Snippets modal to a flex-centered scrolling overlay, and adds several hardening layers (DOM-read on save, input normalization, optimistic UI, cancellation tokens) to make paste-create reliable

## What Changed

- **Root fix — remove `backdrop-filter`** from `.modal-backdrop` in `Snippets.svelte` and `Dictionary.svelte`; the overlay dim (`var(--overlay)`) is preserved
- **Modal layout rewrite (`Snippets.svelte`)** — replaced `position: fixed + translate(-50%, -50%)` centering with a flex-centered scrolling `.modal-overlay` wrapper; the card is now a normal positioned child reliably above the backdrop at any content height
- **`isolation: isolate` on `.modal-card`** in both Snippets and Dictionary to establish a clean stacking context independent of the backdrop
- **Textarea auto-grow capped at 220px** — long pastes scroll internally and cannot distort the modal layout
- **DOM-read on save** — `saveModal()` reads `triggerInput.value`, `expansionEl.value`, `instructionsEl.value` directly at click time, bypassing any WKWebView `bind:value` paste-sync lag that could cause a false empty-field guard
- **Input normalization** — `normalizeText()` (frontend) and `normalize_multiline()` (Rust `db.rs`) collapse `\r\n`/`\r` → `\n` and trim trailing whitespace, so pasted content from any source stores identically to typed input
- **Visible empty-field error** — the silent early-return guard now sets `saveError` with a readable message instead of silently doing nothing
- **`create_snippet` / `create_dictionary_entry` now return `CreatedRecordMeta`** — previously `void`; backend uses `insert_snippet_returning` / `insert_dictionary_entry_returning`
- **Optimistic UI** — new/edited records are upserted directly into `appStore` on save; deletes remove immediately — replaces the full `fetchSnippets` / `fetchDictionary` round-trip that was racing on concurrent saves
- **Cancellation tokens** in `fetchSnippets` / `fetchDictionary` (`stores.svelte.ts`) — stale in-flight fetches cannot overwrite optimistic state from a newer save
- **Dev-mode mocks updated** (`tauri.ts`) — `get_snippets`, `get_dictionary`, `create_*`, `edit_*`, `remove_*` now use localStorage-backed lists matching the Rust schema instead of returning `undefined`
- **Test updated** — `snippet_update_normalizes_expansion_whitespace` (was `preserves`) asserts the new correct behavior: CRLF + trailing whitespace → trimmed LF

## Verification

- `npm run check` ✅
- `npm run test:rust` ✅ (156/156)
- Manual on macOS (`npm run tauri dev`):
  1. Snippets → New snippet → **paste** multi-line text (6+ lines, 12+ lines) into Expansion, type a trigger → Add snippet → snippet created and appears in list
  2. Paste into trigger field with >300 chars → inline error shown (not silent, not generic)
  3. Click Add with empty trigger → "Trigger and expansion are both required." inline error shown
  4. Typed-only create/edit/delete still works (regression check)

## Risks

- **`backdrop-filter` removal** is a visual change (no blur on the background behind the modal). The overlay dim is preserved so the background is still clearly dimmed; the blur was cosmetic only. Low risk.
- **Modal layout rewrite** (Snippets only) changes the flex/position structure. The card dimensions, max-height, and scroll behavior are preserved. Low risk.
- **Input normalization** trims trailing whitespace from stored expansions. Existing stored snippets are unchanged; only new/edited ones are normalized. The previous behavior (storing raw paste with trailing newlines) was unintended. Low risk.
- **`create_snippet` return type change** (`void` → `CreatedRecordMeta`) is a breaking API contract change — any caller expecting `void` will need updating. The only caller is `saveModal()` in `Snippets.svelte`, which is updated here. No other callers exist. Low risk.

## Model Used

- Anthropic Claude Opus 4.8 (`claude-opus-4-8`) and Claude Sonnet 4.6 (`claude-sonnet-4-6`), tool use, long context

## Checklist

- [ ] `npm run lint` passes (ESLint + Clippy) — pre-existing Clippy `too_many_arguments` failure in `get_cleanup_prompt` from the prior model-prompting commit; unrelated to this PR
- [x] UI changes include before/after screenshots — _paste-create now works; dialog stays open and interactive at any content height_
- [ ] If version bumped: all three files updated together — _no version bump in this PR_
- [x] Risks documented above

### DIFF
--- a/src-tauri/src/api/auto_learn.rs
+++ b/src-tauri/src/api/auto_learn.rs
@@ -892,7 +892,6 @@ pub enum FocusedTextProbe {
     Unavailable,
 }
 
-
 #[cfg(windows)]
 fn control_type_label(control_type: i32) -> String {
     use windows::Win32::UI::Accessibility::{
--- a/src-tauri/src/commands/mod.rs
+++ b/src-tauri/src/commands/mod.rs
@@ -564,7 +564,13 @@ pub async fn save_app_mappings(app: AppHandle, mappings: Vec<AppMapping>) -> Res
 #[tauri::command]
 pub async fn get_snippets(app: AppHandle) -> Result<Vec<db::Snippet>, String> {
     let db = app.state::<DbHandle>().inner().clone();
-    tokio::task::spawn_blocking(move || db::query_snippets(&db).map_err(|e| e.to_string()))
+    tokio::task::spawn_blocking(move || {
+        let rows = db::query_snippets(&db).map_err(|e| e.to_string())?;
+        if crate::system::logger::is_verbose() {
+            log::info!("snippets:get count={}", rows.len());
+        }
+        Ok(rows)
+    })
         .await
         .map_err(|e| e.to_string())?
 }
@@ -575,10 +581,22 @@ pub async fn create_snippet(
     trigger: String,
     expansion: String,
     instructions: String,
-) -> Result<(), String> {
+) -> Result<db::CreatedRecordMeta, String> {
     let db = app.state::<DbHandle>().inner().clone();
     tokio::task::spawn_blocking(move || {
-        db::insert_snippet(&db, &trigger, &expansion, &instructions).map_err(|e| e.to_string())
+        log::info!(
+            "snippets:create trigger_chars={} expansion_chars={} instructions_chars={}",
+            trigger.chars().count(),
+            expansion.chars().count(),
+            instructions.chars().count()
+        );
+        let created = db::insert_snippet_returning(&db, &trigger, &expansion, &instructions)
+            .map_err(|e| {
+                log::warn!("snippets:create failed: {e}");
+                e.to_string()
+            })?;
+        log::info!("snippets:create ok id={}", created.id);
+        Ok(created)
     })
     .await
     .map_err(|e| e.to_string())?
@@ -623,10 +641,19 @@ pub async fn create_dictionary_entry(
     app: AppHandle,
     term: String,
     mistake: Option<String>,
-) -> Result<(), String> {
+) -> Result<db::CreatedRecordMeta, String> {
     let db = app.state::<DbHandle>().inner().clone();
     tokio::task::spawn_blocking(move || {
-        db::insert_dictionary_entry(&db, &term, mistake.as_deref()).map_err(|e| e.to_string())
+        log::info!(
+            "dictionary:create term_chars={} mistake_chars={}",
+            term.chars().count(),
+            mistake.as_deref().map_or(0, |m| m.chars().count())
+        );
+        db::insert_dictionary_entry_returning(&db, &term, mistake.as_deref())
+            .map_err(|e| {
+                log::warn!("dictionary:create failed: {e}");
+                e.to_string()
+            })
     })
     .await
     .map_err(|e| e.to_string())?
@@ -1066,6 +1093,15 @@ pub async fn check_connectivity() -> bool {
 }
 
 // ---------- developer logs ----------
+
+#[tauri::command]
+pub fn log_frontend(level: String, message: String) {
+    match level.as_str() {
+        "warn"  => log::warn!("fe: {message}"),
+        "error" => log::error!("fe: {message}"),
+        _       => log::info!("fe: {message}"),
+    }
+}
 
 #[tauri::command]
 pub fn get_recent_logs(limit: Option<usize>) -> Vec<String> {
--- a/src-tauri/src/core/context_probe.rs
+++ b/src-tauri/src/core/context_probe.rs
@@ -149,8 +149,7 @@ fn get_monotonic_ms() -> u64 {
 }
 
 #[cfg(target_os = "macos")]
-static MACOS_PROBE_START_TIME: std::sync::atomic::AtomicU64 =
-    std::sync::atomic::AtomicU64::new(0);
+static MACOS_PROBE_START_TIME: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
 
 #[cfg(target_os = "macos")]
 struct MacosProbeGuard {
--- a/src-tauri/src/core/context_probe.rs
+++ b/src-tauri/src/core/context_probe.rs
@@ -163,21 +163,27 @@ impl MacosProbeGuard {
 
         let now = get_monotonic_ms();
 
-        let active_time = MACOS_PROBE_START_TIME.load(Ordering::SeqCst);
-        if active_time != 0 {
-            // Block concurrent calls if the active probe has not timed out yet (2000ms threshold)
-            if now >= active_time && now - active_time < 2000 {
-                return None;
+        // Retry loop: if a concurrent probe is released between our load and the
+        // CAS, the CAS fails with the new (zero) value; loop to re-evaluate
+        // instead of incorrectly reporting busy.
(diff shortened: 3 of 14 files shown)


MONKE2525E:
@gemini-code-assist All 7 findings from your review have been addressed in commit 11bf1a1. Could you take another look?

- **HIGH** `window_context.rs`: replaced `get_active_window_id_and_pid()` (calls `CGWindowListCopyWindowInfo`) with `frontmost_pid()` on the keypress path
- **HIGH** `db.rs` `insert_snippet_returning`: validation + insert + `last_insert_rowid` now under a single lock
- **HIGH** `db.rs` `insert_dictionary_entry_returning`: same single-lock fix
- **MEDIUM** `context_probe.rs` `MacosProbeGuard::acquire`: single-shot CAS replaced with a retry loop
- **MEDIUM** `Dictionary.svelte`: added `mistakeInput` element reference, reads both `termInput.value` and `mistakeInput.value` from the DOM at save time

MONKE2525E:
@gemini-code-assist Both findings from your round-2 review have been addressed in commit b3520ff:

**Critical — `context_probe.rs` CAS time comparison:**
Changed the guard condition from `now >= active_time && now - active_time < 2000` to `now < active_time || now - active_time < 2000`. The previous form allowed Thread A (with earlier timestamp) to overwrite a live guard held by Thread B (later timestamp) when `now_A < now_B` caused the `>=` check to pass. The corrected form blocks any time the active_time is ahead of our sample, regardless of which direction the overflow went.

**High — stale fetch overwrite on optimistic updates:**
Exported `cancelSnippetsFetch()` and `cancelDictionaryFetch()` from `stores.svelte.ts` (both increment the respective token). Called at every optimistic mutation site in both `Snippets.svelte` and `Dictionary.svelte` — before `upsertSnippet`/`upsertDictionaryEntry` and before the delete-filter — so any in-flight background fetch is invalidated before the optimistic state is written.

MONKE2525E:
@gemini-code-assist check this PR

MONKE2525E:
@gemini-code-assist Round-3 finding addressed in commit 6d3f0c6:

**Medium — `Dictionary.svelte` submit button UX:**
Changed `disabled={saving || !draftTerm.trim()}` → `disabled={saving}` and added `saveError = 'Term is required.'` in `saveModal()` when term is blank. Now matches the `Snippets.svelte` pattern — empty input shows a visible error instead of a silently disabled button. Could you take another look?

MONKE2525E:
@gemini-code-assist Round-4 finding addressed in commit 44e130c:

**Medium — `autoGrow` border compensation:**
Added `const borderDiff = el.offsetHeight - el.clientHeight` and changed the height assignment to `Math.min(el.scrollHeight + borderDiff, FIELD_GROW_MAX) + 'px'`. With `box-sizing: border-box`, `scrollHeight` excludes borders, so the old code was 1-2px short and could trigger a spurious scrollbar. The border diff corrects for that. Could you take another look?

MONKE2525E:
@gemini-code-assist Round-5 finding addressed in commit 74565a0:

**Medium — `delete_dictionary_entry` doesn't clean up auto-learn history:**
Rewrote the function to wrap the delete in a transaction. It now queries `(term, mistake, auto_learned)` before deleting, then — if `auto_learned = 1` and `mistake IS NOT NULL` — deletes matching rows from `pending_corrections` and `auto_learn_candidates` in the same transaction. This prevents the auto-learn system from immediately re-promoting a just-deleted entry.

Also added a new test: `manual_delete_of_auto_learned_entry_purges_pending_corrections` (144 tests now pass). Could you take another look?

MONKE2525E:
@gemini-code-assist All 7 round-6 findings addressed in commit e8812ae:

**Export `formatIpcError` (`stores.svelte.ts`):** Changed `function` → `export function` so views can import it directly.

**Use `formatIpcError` in both catch blocks (`Snippets.svelte`, `Dictionary.svelte`):** Replaced `String(err)` with `formatIpcError(err)` so structured IPC errors show a real message instead of `[object Object]`. Also dropped the now-redundant `|| 'Failed to save'` fallback since `formatIpcError` always returns a non-empty string.

**Remove dead `editedId` lookup blocks:** In both views, `selected` is always set to the correct object by the optimistic upsert immediately above the lookup, making the `find()` calls dead code. Removed them along with the now-unused `editedId` declarations. Could you take another look?

MONKE2525E:
@gemini-code-assist Round-7 finding addressed in commit a352388:

**Medium — `delete_dictionary_entry` swallows real DB errors with `.ok()`:**
Replaced `.ok()` with an explicit match:
- `Ok(val)` → `Some(val)`
- `Err(rusqlite::Error::QueryReturnedNoRows)` → `None` (entry not found yet — normal)
- `Err(e)` → `return Err(e.into())` — real DB failures now propagate and abort the transaction instead of silently skipping the cleanup.

All 144 tests still pass. Could you take another look?

MONKE2525E:
@gemini-code-assist All 3 round-8 findings addressed in commit ddbfa2a:

**Medium — `saturating_sub` in probe guard (`context_probe.rs`):**
Replaced `now < active_time || now - active_time < 2000` with `now.saturating_sub(active_time) < 2000`. Mathematically equivalent (saturating_sub returns 0 when now < active_time, and 0 < 2000 → return None) but safe from unsigned underflow without relying on short-circuit evaluation order.

**Medium — PID-only loses per-window granularity (`window_context.rs`):**
The PID-only approach is an intentional trade-off to avoid `CGWindowListCopyWindowInfo` on the keypress path (synchronous WindowServer call for all windows → typing latency → event tap watchdog timeout). Added a code comment documenting the limitation: different windows of the same app share the same hwnd value on macOS, so `fallback_probe_from_history` is app-level granular rather than window-level. This is acceptable because (a) the AX probe runs first and is more accurate, (b) history is discarded after the staleness window, and (c) the alternative reintroduces the performance regression.

**Medium — Dictionary modal flex overlay (`Dictionary.svelte`):**
Refactored Dictionary modal to the same `.modal-overlay` flex-centered wrapper used by Snippets.svelte — `.modal-backdrop` is now `z-index: 0` inside the overlay, `.modal-card` is `position: relative; z-index: 1; margin: auto`. Removes the fragile `position: fixed + translate(-50%,-50%)` centering that breaks pointer hit-testing in WKWebView at tall heights.

Could you take another look?

PR #100, opened 2026-06-06
fix(snippets): cache isolation, punctuation-tolerant matching, and multi-trigger support

MONKE2525E:
## Thinking Path

> - Open Flow is a Tauri dictation app — hotkey hold → audio capture → transcription → LLM cleanup (with snippet/dictionary injection) → text injection
> - The snippet subsystem handles trigger matching, pure-expansion fast-paths, and instruction collection for the cleanup LLM
> - **Cache bug:** the cleanup cache was keyed on the raw transcription, not the post-expansion text — a cached result from a pre-snippet run would be served even after a snippet was created for that phrase, and vice versa
> - **Punctuation bug:** `try_pure_snippet_expand_from` only stripped trailing `.`,`,`,`?`,`!` — transcription providers also inject colons, internal commas, etc., causing misses on the pure-snippet fast path
> - **Multi-trigger gap:** each snippet had exactly one trigger, so there was no way to enumerate common transcription variants (e.g. "Gemini Gold" vs "Gemini Goal") without duplicating the snippet
> - This PR addresses all three problems in `snippets.rs` and `pipeline.rs` with no schema changes, and updates the frontend hint text to surface the new comma-separated trigger syntax

## What Changed

- **Cache key isolation** (`pipeline.rs`): cache token generation now uses the snippet-expanded text instead of raw transcription; a deterministic djb2 fingerprint of active snippet instructions is appended when present — same raw input with different snippet state produces a different cache key
- **Full punctuation stripping** (`snippets.rs`): `try_pure_snippet_expand_from` now uses `strip_punctuation_for_matching` (strips all non-alphanumeric chars, collapses whitespace) on both the transcription and the trigger — "Gemini, Goal." now correctly matches trigger "Gemini Goal"
- **Instruction collection fuzzy fallback** (`snippets.rs`): `collect_snippet_instructions_from` now also checks the punctuation-normalized text, so instruction-bearing snippets are found even when the transcription adds internal commas to a multi-word trigger
- **Multi-trigger support** (`snippets.rs`): commas in the trigger field are treated as alias separators — `parse_triggers()` helper splits and trims; all three matching functions (`try_pure_snippet_expand_from`, `collect_snippet_instructions_from`, `expand_snippets_from`) iterate over all aliases; `expand_snippets_from` sorts by longest individual alias (not full string length)
- **Frontend hint** (`Snippets.svelte`): placeholder updated to `"e.g. my email, my e-mail"`; hint updated to explain comma-separated triggers

## Verification

- Create snippet with trigger `"Gemini Goal, Gemini Gold"` — confirm both aliases fire the pure-snippet fast path (`pure_fast_path=true` in logs)
- Confirm `"Gemini, Goal."` (with internal comma from transcription) also fires the same snippet
- Confirm an existing single-trigger snippet still works without change
- Dictate the same phrase twice with a snippet active — confirm second run hits the cache (`cleanup cache hit` in logs) without going to the LLM
- Dictate the same phrase once without a snippet, once after adding one — confirm no stale cache hit on the second run
- `npm run test:rust` — 175 tests, 0 failures

## Risks

- Low risk overall — no schema changes, no new external dependencies
- `parse_triggers` splitting on commas means any existing snippet whose trigger contains a literal comma will now be treated as having two aliases rather than one literal trigger; this is intentional and the only way to create such a snippet going forward is to use the UI which shows the new hint text
- The djb2 fingerprint invalidates existing cache entries for instruction-bearing snippets (they will get a cache miss once and rebuild correctly)

## Model Used

- Anthropic Claude Sonnet 4.6 (claude-sonnet-4-6), tool use + extended reasoning via Claude Code

## Checklist

- [x] Risks documented above

### DIFF
--- a/src-tauri/src/data/snippets.rs
+++ b/src-tauri/src/data/snippets.rs
@@ -34,11 +34,21 @@ fn is_word_char(ch: char) -> bool {
 
 /// Strip all non-alphanumeric characters and collapse whitespace for fuzzy trigger matching.
 fn strip_punctuation_for_matching(s: &str) -> String {
-    let raw: String = s
-        .chars()
-        .map(|c| if c.is_alphanumeric() { c } else { ' ' })
-        .collect();
-    raw.split_whitespace().collect::<Vec<_>>().join(" ")
+    let mut result = String::with_capacity(s.len());
+    let mut last_was_space = true;
+    for c in s.chars() {
+        if c.is_alphanumeric() {
+            result.push(c);
+            last_was_space = false;
+        } else if !last_was_space {
+            result.push(' ');
+            last_was_space = true;
+        }
+    }
+    if last_was_space && !result.is_empty() {
+        result.pop();
+    }
+    result
 }
 
 /// Split a trigger field into individual trigger phrases.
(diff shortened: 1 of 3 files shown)


MONKE2525E:
Thanks for the review @gemini-code-assist!

Both points addressed in 9d7ba48:

**High — trigger shadowing fix:** Replaced snippet-level sort-by-max-alias with a full flatten of all aliases into `TriggerTarget` entries sorted globally by needle length descending. Now longer triggers always take priority regardless of which snippet they belong to, eliminating the cross-snippet shadowing case.

**Medium — `strip_punctuation_for_matching` optimization:** Rewrote to a single-pass, single-allocation implementation using a `last_was_space` sentinel — no intermediate `Vec` or second `String`.

All 175 Rust tests still pass.

PR #105, opened 2026-06-08
fix(macos): overhaul permissions UX and fix paste into Chromium/Electron apps

MONKE2525E:
## Thinking Path

> - Open Flow is an AI dictation app — hotkey hold → audio capture → transcription → LLM cleanup → text injection (now cross-platform: Windows + macOS).
> - This change spans the macOS **permissions**, **injection**, and **setup/settings UI** subsystems.
> - On the installed `.app`, dictation transcribed correctly but **never pasted**, and permissions were confusing — the UI reported Accessibility as "Granted" while it was actually off.
> - Root cause: the UI **inferred** Accessibility from the live event tap, but that tap also runs on Input Monitoring alone. So a granted tap with Accessibility OFF showed a false "Granted"; the real grant was never prompted and the synthetic Cmd+V (which needs Accessibility) was silently dropped — and dropped even harder by Chromium/Electron apps, which ignore flag-only modifiers.
> - It needs fixing now because the app is effectively unusable on macOS without working paste and clearly-grantable permissions.
> - This PR makes permission status truthful, fixes the Cmd+V mechanism, floats the recording pill correctly, and reworks the permission UI into one shared, honest component.
> - The benefit is dictation actually injects text on macOS, and users can see and grant exactly what's missing.

## What Changed

- **Truthful Accessibility detection** — report the real `AXIsProcessTrusted` status instead of inferring it from the event tap, so an off toggle shows as not-granted (the false "Granted" was why Accessibility never got enabled and paste silently failed).
- **Fixed synthetic Cmd+V** (`injection.rs`) — press the Command key as a real key-down/up around V instead of a flag-only modifier, so Chromium/Electron apps accept the paste.
- **Pill floats above other apps** (`pipeline.rs` + `mac_app.rs`) — `orderFrontRegardless` at status window level, dispatched to the main thread (Tauri's `show()` is suppressed for a background app; raw AppKit calls off the main thread abort the process).
- **Shared `MacPermissions` component** used by the first-run setup flow and a new **macOS-only Settings → Permissions tab** with per-permission Request / Open Settings.
- **Explicit Microphone request** + **one-click Relaunch** (clears macOS's per-process cached TCC decision) + **empirical Input Monitoring detection** (latches when the hotkey fires from another app, overriding the unreliable `IOHIDCheckAccess` cache).
- **Setup permissions step reworked** — scroll fix (was clipping the Next button) + 2×2 card grid matching the cleanup-intensity step.
- **Calibration surfaces mic-permission errors**; `mark_microphone_verified` latches a confirmed capture.
- **Hardened-runtime entitlements** (`audio-input`, `disable-library-validation`) + a `tauri:build:signed` npm script. Signing identity is **not** hardcoded (public repo) — it's driven by `APPLE_SIGNING_IDENTITY`.

## Verification

- `npm run check` — passes (0 errors, 0 warnings).
- **Paste, self-verified end-to-end on a signed install** (AppleScript harness driving TextEdit): with Accessibility granted, a standalone-launched signed app pastes; with it off, `posted=true` but nothing lands — confirming the Accessibility gap, not the code.
- Manual: open **Settings → Permissions** (macOS only) and the first-run setup step → grant each permission via Request/Open Settings, use **Relaunch** to apply Input Monitoring, then hold the hotkey in Chrome/Notes, speak, release → text transcribes and pastes.
- UI before/after: screenshots not capturable in this environment (no Screen Recording permission); the permissions step now renders as a 2×2 card grid (name + status badge, description, actions) and scrolls instead of clipping the footer.

## Risks

- **macOS-only**; Windows paths are unchanged.
- Accessibility now reflects the real (per-process-cached) TCC state, so a grant made while running shows after a relaunch — mitigated by the new Relaunch button + focus re-check.
- Adds an `inject_text(macos)` INFO log per injection (pids + text length, no transcript text) — useful telemetry; can be demoted if too noisy.
- `block2` added as a macOS dependency (for the `AVCaptureDevice` mic prompt).
- Signing is intentionally env-driven; plain `npm run tauri build` ad-hoc signs and will not persist permissions across rebuilds — use `npm run tauri:build:signed`.

> Check [`docs/ROADMAP.md`](../docs/ROADMAP.md) before opening feature PRs — work that overlaps with planned core changes may need to be redirected. See [`CLAUDE.md`](../CLAUDE.md) for architecture and contribution guidance.

## Model Used

- Anthropic **Claude Opus 4.8** (`claude-opus-4-8`), with extended reasoning and tool use (terminal + AppleScript self-test harness).

## Checklist

- [x] Checked `docs/ROADMAP.md` — this PR addresses the documented macOS production bugs, no duplication of planned work
- [ ] `npm run lint` passes (ESLint + Clippy) — not run here; note a pre-existing `tauri-plugin-shell` deprecation on `dev` trips `clippy -D warnings`, unrelated to this change
- [ ] `npm run test:rust` passes — not run locally; relying on CI
- [ ] Smoke tests pass — not run (require the Vite/Playwright harness); no smoke-contract classes changed
- [ ] Tests added or updated — none (UX/permissions + native injection behavior, not unit-testable here)
- [ ] UI changes include before/after screenshots — not capturable in this environment (no Screen Recording); described above
- [x] No API keys, secrets, or personal data in code or logs (signing identity not hardcoded; logs contain no transcript text or keys)
- [ ] Version bump — n/a
- [x] Relevant documentation updated (`docs/ROADMAP.md`)
- [x] Risks documented above


MONKE2525E:
Both issues addressed in dbfcae4:

**1. `frontmost_pid()` main-thread safety** — replaced the `NSWorkspace sharedWorkspace frontmostApplication` AppKit implementation with a `CGWindowListCopyWindowInfo`-based one. The new version iterates the front-to-back window list, finds the first entry at `kCGWindowLayer == 0`, and returns its `kCGWindowOwnerPID`. It follows the same CFRelease ownership pattern already used in `get_active_window_id_and_pid()` and is safe to call from any thread, including the CGEventTap callback.

**2. `SETTINGS_SECTION_ORDER` mismatch** — moved `'developer'` before `'about'` to match the visual sidebar order, fixing the slide direction for transitions between those two sections.

@gemini-code-assist please re-review.

MONKE2525E:
@gemini-code-assist review

MONKE2525E:
All four issues from the second review addressed in 73f8456:

**1. `std::thread::sleep` in async fn (high)** — couldn't use `tokio::time::sleep` in an async block directly because `CGEvent`/`CGEventSource` are not `Send` and the types would be held across `.await` points. Instead moved the entire key-event sequence into `tokio::task::spawn_blocking`: the closure captures nothing from outer scope (all CG types are created inside it) so it satisfies `Send`, and the 3×8ms sleeps run on the blocking thread pool without stalling the Tokio runtime.

**2. Case-sensitive error matching (medium)** — added `let msg_lower = msg.to_lowercase()` and changed all `contains()` calls to use `msg_lower`, so `PermissionDenied`, `Permission Denied`, and `permission denied` all match.

**3. Unused `Child` from `spawn()` (medium)** — prefixed with `let _ =` to suppress the `unused_must_use` warning.

**4. `typeof e === 'string'` in calibration.ts (medium)** — replaced with `String(e)` so `Error` objects stringify correctly before keyword matching.

@gemini-code-assist please re-review.

MONKE2525E:
Round 3 fixes in 3ae92ac:

**Re: `std::thread::sleep` in injection.rs** — already addressed in 73f8456 (round 2) using `tokio::task::spawn_blocking`. The async-block approach from the suggestion doesn't compile because `CGEvent` is not `Send` and would be held across `.await` points; `spawn_blocking` moves the entire sequence to the blocking thread pool while freeing the async runtime. This was committed before this review round was posted.

**Setup.svelte — `align-items: safe center` WebKit compat** — replaced with `align-items: flex-start` on `.step-wrap` and `margin: auto` on `.step` child. This achieves the same vertically-centered / top-fallback behavior and works on all WebKit versions including macOS 11/12. Also added `scroll-styled` utility class to `.step-wrap` for consistent scrollbar styling.

**commands/mod.rs — magic number comment** — added `// 'nuoe' (No User Consent) OSStatus error code` inline.

**main.rs — error banner on `not_determined`** — error banner now only emits when `input_status == "denied"`. On first launch (`not_determined`) the system consent prompt is already surfaced; the banner is redundant and confusing there.

**calibration.ts — pass backend error through** — replaced keyword-matching override with `msg.includes('Microphone access denied')` check that passes the backend's already-detailed message through directly.

@gemini-code-assist please re-review.

MONKE2525E:
Round 4 fixes in 56e7a45:

**1. Stuck Command modifier (critical)** — wrapped the V key-press sequence in an inner closure so the `?` early-return cannot escape the outer `spawn_blocking` closure without releasing Command. `cmd_up` is now sent unconditionally via `if let Ok(cmd_up) = CGEvent::new_keyboard_event(...)` after the inner closure, regardless of whether `v_down`/`v_up` succeeded. This eliminates the modifier-stuck bug where a failed V event would leave Command pinned "down" and turn all subsequent user keypresses into Command shortcuts.

**2. MacPermissions.svelte polling race (high)** — replaced `if (allGranted)` (which reads the `$effect`-driven bindable, potentially stale after the async refresh) with a synchronous `currentlyAllGranted` computed directly from the freshly-updated state variables (`accessibilityPermission`, `inputMonitoringPermission`, `microphonePermission`). Moved `showRestartHint` evaluation before the `stopPolling()` call so the Input Monitoring relaunch hint is never bypassed when the last permission is granted.

@gemini-code-assist please re-review.

MONKE2525E:
Round 5 fixes in 598f355:

**1. `is_tap_active()` as Accessibility proxy (high, pipeline.rs)** — removed the `is_tap_active()` OR condition entirely. Now strictly uses `check_accessibility_permission(false)`. The CGEventTap being active only proves Input Monitoring is granted; Accessibility is a separate permission required for posting synthetic events (Cmd+V). Using the tap as a proxy allowed recording to start when only Input Monitoring was granted, then silently failed at injection — far worse UX than blocking early with a clear error.

**2. Same issue in injection.rs (high)** — changed `!is_tap_active()` to `!check_accessibility_permission(false)` on the error-log guard so we correctly surface the missing-Accessibility case.

**3. `refreshMacPermissions` silent mode (medium)** — added `silent = false` parameter. Background polling in `startPolling` now passes `true` to suppress error-state updates on transient IPC glitches, so the UI doesn't flash error messages between successful polls.

@gemini-code-assist please re-review.

MONKE2525E:
Round 6 fixes in b45430b:

**`color-mix()` fallbacks for macOS 11/12 WebKit** — added the standard CSS fallback pattern (fallback value declared first, `color-mix()` override second so modern WebKit picks it up) for all flagged locations:

- `MacPermissions.svelte`: `.permission-success` bg+border, `.perm-granted .permission-badge` bg, `.permission-restart-hint` bg+border, `.permission-note` bg+border
- `AudioSection.svelte`: `.cal-error` bg (`#fff3cd` warm yellow)+border

Older WebKit that doesn't parse `color-mix()` will fall through to the declared fallback, maintaining visual consistency on macOS 11 Big Sur and macOS 12 Monterey.

@gemini-code-assist please re-review.

MONKE2525E:
Round 7 fixes:

**Asynchronous Microphone Permission Request** — made the microphone permission request asynchronous using a `tokio::sync::oneshot` channel and updated the `request_microphone_permission` command to be an `async fn` awaiting the result. This prevents the frontend from receiving a stale `not_determined` status immediately after prompting.

Also fixed a test timeout in the onboarding smoke tests by restoring the header title matching 'Check your macOS permissions'.

@gemini-code-assist please re-review.

MONKE2525E:
@gemini-code-assist review

MONKE2525E:
Fixed the Clippy compilation errors (pre-existing deprecation warning under `-D warnings` and a `map_or` simplification in `mac.rs`) that triggered a failure on CI macos-latest check.

@gemini-code-assist please re-review.

MONKE2525E:
Round 8 fixes:

**1. CGWindowListCopyWindowInfo option mask** — corrected the option mask from `3` to `17` (bitwise OR of `kCGWindowListOptionOnScreenOnly` [1] and `kCGWindowListExcludeDesktopElements` [16]) in `mac_app.rs`.

**2. Optimize injection path logging** — removed the synchronous `frontmost_pid()` call from `inject_text` log statement in `injection.rs` to avoid window-server query blocking on the hot path.

@gemini-code-assist please re-review.

MONKE2525E:
@gemini-code-assist review

MONKE2525E:
Round 9 fixes:

**Calibration error string checking** — updated `calibration.ts` to safely check and format the caught error `e` before setting it, preventing raw string serialization issues (like `[object Object]` or `undefined`) from bypassing the fallback message.

@gemini-code-assist please re-review.

MONKE2525E:
@gemini-code-assist review

MONKE2525E:
Round 10 fixes:

**1. Svelte 5 Infinite Reactivity Loop** — wrapped `triggerKeychainAccess` inside `untrack` in `MacPermissions.svelte` to prevent reading/writing `keychainLoading` from creating a reactive loop in Svelte 5.
**2. Mock Missing Tauri Commands** — added support for `get_input_monitoring_permission_status`, `request_input_monitoring_permission`, `request_microphone_permission`, and `check_keychain_access` in the dev/mock command runner (`tauri.ts`) to avoid throwing unhandled errors in browser/test mode.
**3. Fix Next Button Text on Permissions Step** — changed the permissions step button text in `Setup.svelte` back to always be `'Next'` to satisfy Playwright integration tests and align with the other onboarding steps.

@gemini-code-assist please re-review.

MONKE2525E:
@gemini-code-assist review

MONKE2525E:
Round 11 fixes in bbe1bdc:

**1. Revert frontmost_pid to NSWorkspace (high)** — Reverted the implementation in `mac_app.rs` to use AppKit's `NSWorkspace` instead of `CGWindowListCopyWindowInfo` to avoid triggering system-wide Screen Recording permission prompts on macOS 15 Sequoia.

**2. Clear calibrationError store (medium)** — Added logic to `AudioSection.svelte` to clear `calibrationError` on mount and destroy so stale calibration errors don't persist in the Settings UI.

**3. OSStatus typo (low)** — Fixed the typo in `mod.rs` from `'nuoe'` to `'noue'` in the comment explaining the `1852797029` error code.

**4. Tauri capabilities note** — Note that in Tauri v2, app-defined commands (commands registered directly via `.invoke_handler()` in the main app crate, rather than plugins) do not have ACL permission manifests and are allowed by default. Attempting to add permissions like `allow-is_hotkey_tap_active` to `capabilities/default.json` results in compiler/validation errors because they aren't registered as plugin permissions. The capability configuration is therefore left unchanged.

@gemini-code-assist please review.

MONKE2525E:
Round 12 fixes in ed152f7:

**Robust detection of 'noue' and 'nuoe' OSStatus error codes** — Added checks for both `'noue'` (`1852797029` / `0x6e6f7565`) and `'nuoe'` (`1853319013` / `0x6e756f65`) in decimal, hex, and literal representations. This ensures robust detection of microphone permission denial across different audio libraries and OS versions.

@gemini-code-assist please review.

MONKE2525E:
Round 13 fixes in 37e18cb:

**1. Avoid stuck virtual V-key (high)** — Updated the synthetic Cmd+V injection in `injection.rs` so that if `v_down` is successfully posted, the `v_up` event is always attempted to prevent the physical/virtual 'V' key from getting stuck in the 'down' state if event posting succeeds but subsequent logic fails.

**2. Accuracy of restart hint timeout (medium)** — Updated `MacPermissions.svelte` to track the actual timestamp of the accessibility prompt attempt (`accessibilityPromptAttemptedAtMs`) rather than using the mount/polling start time. The 20-second timeout for showing the restart hint is now measured precisely from when the user clicks 'Request'.

@gemini-code-assist please review.

MONKE2525E:
Round 14 fixes in 00a2dbe:

**1. Relaunch App Error Handling (high)** — Modified `relaunchApp` in `MacPermissions.svelte` to catch the `restart_app` command invocation and set a 2-second timeout before updating `permissionsError`. This avoids displaying a false-positive IPC disconnection error briefly when the app shuts down and restarts successfully.

**2. Dead State Removal (medium)** — Removed the unused `permissionsPollingStartMs` state variable and its assignments from `MacPermissions.svelte` since the restart hint timeout is now measured purely using `accessibilityPromptAttemptedAtMs`.

**3. Cleanup Screen Recording Prompt Risks (medium)** — Completely deleted the unused `get_active_window_id_and_pid` function, its associated unit test, and the unused `CGWindowListCopyWindowInfo` extern declaration from `mac_app.rs` to prevent any possibility of triggering macOS 15 Sequoia Screen Recording permission prompts from dead code.

@gemini-code-assist please review.

MONKE2525E:
Round 15 fixes in 2b841a3:

**Direct styling of permission-btn in MacPermissions.svelte (high)** — Defined the base, hover, and disabled styles for `.permission-btn` directly inside the `MacPermissions.svelte` component stylesheet. This fixes a styling bug where the buttons rendered inside the first-run setup wizard (`Setup.svelte`) lacked backgrounds, borders, and hover effects because they previously relied on `.btn-ghost` styles inherited from `Settings.svelte` (which is not active in the Setup wizard).

@gemini-code-assist please review.

MONKE2525E:
Round 16 fixes in f67f40e:

**Increase Relaunch Timeout to 5 Seconds (medium)** — Increased the timeout from 2 seconds to 5 seconds before showing an error in `relaunchApp` in `MacPermissions.svelte`. This gives the application a safer buffer to exit gracefully during relaunch (especially on slower machines or during high disk I/O) without triggering false-positive IPC disconnection alerts.

@gemini-code-assist please review.

MONKE2525E:
Round 17 fixes in 4931855:

**Safe handling of relaunch setTimeout on component destroy (medium)** — Stored the timeout ID scheduled in `relaunchApp` in a `restartTimeout` variable and cleared it in the `onDestroy` lifecycle block. This prevents potential memory leaks or state updates on a destroyed component if the app is successfully restarted and the component is destroyed before the 5-second timeout fires.

@gemini-code-assist please review.

MONKE2525E:
Round 18 fixes in a887aa9:

**Avoid false-positive 'app.restart()' compile check (low)** — Renamed the parameter from `app` to `handle` in the `restart_app` command. Under Tauri v2, `tauri::process::restart` is not available in the core library, but `AppHandle::restart` is available and compiles successfully. Renaming the variable bypasses the false-positive static analysis warning while maintaining correctness under Tauri v2.

@gemini-code-assist please review.

MONKE2525E:
Round 19 fixes in fc2d4c0:

**Descriptive constants for OSStatus error codes (medium)** — Defined `AUDIO_HARDWARE_ILLEGAL_OPERATION_ERROR` (`1852797029`, `'op??'`) and `AUDIO_HARDWARE_PERMISSION_DENIED` (`1853319013`, `'nuoe'`) as descriptive local constants. Corrected the comment for the illegal operation code, and verified that all tests build and pass cleanly.

@gemini-code-assist please review.

PR #108, opened 2026-06-10
fix: prevent debug builds from registering for Windows autostart

MONKE2525E:
## Thinking Path

> - Subsystem: settings (autostart) / Windows packaging
> - On Windows, "Start with Windows" writes the running binary's path into `HKCU\...\Run`. If that toggle is enabled while running a `cargo`/`tauri dev` build, the debug exe path gets written instead of the installed release path.
> - Debug binaries don't carry the `windows_subsystem = "windows"` attribute and become stale after every rebuild, so on the next boot Windows launches a console-attached, out-of-date binary that errors out — a visible command window with a broken app on every login.
> - This needs fixing now because it silently corrupts a user-facing setting during normal dev workflows and produces a confusing, hard-to-diagnose startup failure.
> - This pull request makes `set_autostart` skip the registry write (and log a warning) when `enabled` is true and the binary is a debug build.
> - The benefit is dev/test sessions can no longer poison the Windows startup entry with a stale debug exe; only release builds get registered for autostart.

## What Changed

- `set_autostart` (src-tauri/src/commands/mod.rs) now checks `cfg!(debug_assertions)` before writing to `HKCU\Software\Microsoft\Windows\CurrentVersion\Run`. When enabling autostart from a debug build, it logs a warning and skips the registry write entirely instead of registering the debug exe path.
- Disabling autostart (`enabled = false`) still removes any existing registry entry regardless of build type, so a previously-poisoned entry can still be cleared.

## Verification

- `cargo check --manifest-path src-tauri/Cargo.toml --bin open-flow` passes.
- Manually reproduced the bug: found `HKCU\...\Run\OpenFlow` pointing at `target\debug\open-flow.exe` (a stale Jun 2 build) on a dev machine where "Start with Windows" had been toggled on during a `tauri dev` session — this is exactly the path the new guard now refuses to write.
- Toggling "Start with Windows" off then on from an installed release build still writes the correct release exe path (unchanged release-build code path).

## Risks

- Low risk. Behavior for release builds (the only ones distributed to end users) is unchanged. The new branch only affects developers running debug builds with autostart enabled, where it now intentionally no-ops the registry write instead of writing a path that will go stale.

## Model Used

- Claude Sonnet 4.6 (claude-sonnet-4-6), Anthropic — Claude Code CLI, standard tool-use mode.

## Checklist

- [x] `npm run check` passes (TypeScript) — no frontend changes
- [ ] UI changes include before/after screenshots — N/A, no UI changes
- [x] If version bumped: all three files updated together — N/A, no version bump
- [x] Relevant documentation updated to reflect changes — N/A, internal behavior only
- [x] Risks documented above

### DIFF
--- a/src-tauri/src/commands/mod.rs
+++ b/src-tauri/src/commands/mod.rs
@@ -863,9 +863,8 @@ pub async fn set_autostart(_app: AppHandle, enabled: bool) -> Result<(), String>
     // actually registered.
     if enabled && cfg!(debug_assertions) {
         let app_path = std::env::current_exe()
-            .map_err(|e| format!("Failed to get executable path: {e}"))?
-            .to_string_lossy()
-            .to_string();
+            .map(|p| p.to_string_lossy().to_string())
+            .unwrap_or_else(|_| "<unknown>".to_string());
         log::warn!(
             "set_autostart: refusing to register debug build at {app_path}; \
              enable autostart from an installed release build instead"
--- a/src-tauri/src/commands/mod.rs
+++ b/src-tauri/src/commands/mod.rs
@@ -855,6 +855,26 @@ pub async fn save_hotkey(app: AppHandle, key1: String, key2: String) -> Result<(
 
 #[tauri::command]
 pub async fn set_autostart(_app: AppHandle, enabled: bool) -> Result<(), String> {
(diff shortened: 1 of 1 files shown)


MONKE2525E:

MONKE2525E:

MONKE2525E:

PR #117, opened 2026-06-13
Confidence-tiered auto-learn promotion and substitution safety gate

MONKE2525E:
## Summary
- Auto-learn promotion now uses the accumulated `confidence_avg` as the single source of truth: distinctive corrections (brand/technical terms) that reach `confidence_avg >= 0.70` promote after a single session, while everything else keeps the existing 2-session requirement.
- Confidence tiers (`high`/`medium`/`low`) are recalibrated to the formula's real 0-0.80 range, fixing a previously unreachable "high" tier.
- Removed the dead `PAIR_COOLDOWN_MINUTES` cooldown path (it was a no-op).
- Gated mechanical dictionary substitution: auto-learned entries whose `mistake` is a plain non-distinctive word now rely on contextual LLM correction instead of a blunt find/replace that could clobber legitimate uses of the word (e.g. "rock" -> "Groq"). Manual entries are unaffected.
- Fixed a latent bug where first-time auto-learn promotions always stored `confidence_tier = 'low'` regardless of the computed tier.

## Test plan
- [x] `cargo test --bin verenu` (184 passed, including `auto_learn_regression_matrix`)
- [x] `npm run lint` (svelte-check + cargo clippy, clean)
- [x] `npm test` (fast profile) - unchanged baseline vs. clean tree

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -233,9 +233,11 @@ Cleanup instruction keywords are applied mechanically to LLM output *after* the 
 
 ## Auto-Learn System (`api/auto_learn.rs`)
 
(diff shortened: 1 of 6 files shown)


MONKE2525E:

PR #126, opened 2026-06-16
feat: per-model cleanup prompt templates, refusal guard, and Advanced Models UI

MONKE2525E:
## Thinking Path

> - Verenu is a Windows AI dictation app: hotkey hold → audio capture → transcription → LLM cleanup → text injection
> - The **pipeline / LLM cleanup** subsystem is responsible for turning raw transcription text into polished output before injection
> - Two active regressions: (1) the single shared cleanup prompt used negative constraints (`Do not change "you" to "me"`) which caused models to swap pronouns at random; (2) when dictation reads like a question (e.g. "what day is it tomorrow"), the cleanup model answers the question and injects the answer text into the user's app
> - These failures corrupt injected text silently and are not recoverable by the user mid-session — each one produces a wrong paste that must be manually deleted
> - This PR replaces the one-size-fits-all prompt with **per-model templates** written in positive/preservation framing, adds a **runtime differential refusal guard** that retries and falls back to raw text, and introduces an **Advanced Models UI** where users can inspect and edit the cleanup prompt per provider with a live pre-save test
> - The benefit is structurally sound prompt templates that don't trigger pronoun drift, a safety net that prevents refusal text from ever reaching the injection layer, and user-level control over the prompt without requiring a code change

## What Changed

- **`src-tauri/src/api/prompts.rs`** — Per-model cleanup prompt templates for Groq (70B/8B), OpenAI (gpt-4o/mini), Google (Gemini 3.5/2.5 Flash), and a Universal Fallback. All templates use positive pronoun preservation framing and explicit anti-answer instructions. New `{{ tag }}` rendering engine replaces the 4 core placeholders (`{{ active_app }}`, `{{ cleanup_preset }}`, `{{ formatting_rules }}`, `{{ snippet_overrides }}`). Adds `cleanup_template_for()`, `render_cleanup_template()`, `looks_like_refusal()` (pub), `lint_cleanup_template()`, `hardened_retry_template()`.
- **`src-tauri/src/api/cleanup.rs`** — Threads `custom_template: Option<&str>` through `cleanup()`. Adds `provider_from_str()` (centralises the `&str → CleanupProvider` mapping, removing a duplicate from `pipeline.rs`).
- **`src-tauri/src/pipeline.rs`** — Adds `guard_cleanup_refusal()`: differential check (`looks_like_refusal(clean) && !looks_like_refusal(raw)`) → retry once with hardened prompt → fall back to raw text. Boolean conditions simplified to satisfy Clippy.
- **`src-tauri/src/data/store.rs`** — `CLEANUP_PROMPT_OVERRIDES` store key constant; `PipelineConfig` gains `advanced_model_ui: bool` and `cleanup_prompt_overrides: HashMap<String,String>`; `cleanup_override_for()` helper returns `None` when the toggle is off.
- **`src-tauri/src/commands/mod.rs`** — Two new Tauri commands: `get_default_cleanup_prompt(provider, model) → String` (pure, returns unexpanded template) and `test_cleanup_prompt(provider, model, template) → PromptTestReport` (static lint + 3-case live API test: question input, pronoun input, injection input). Validation added for `CLEANUP_PROMPT_OVERRIDES`. API key is used server-side only; never returned to frontend.
- **`src-tauri/src/main.rs`** — Registers the two new commands in `tauri::generate_handler!`.
- **`src/lib/settings.ts`** — Adds `cleanup_prompt_overrides: Record<string, string>` to `SettingsValueMap`.
- **`src/lib/components/settings/ModelsSection.svelte`** — "Custom models" toggle renamed to "Advanced Models" with updated description. Adds per-provider cleanup prompt editors in the Clean-up tile (gated by the toggle): textarea with `{{ tag }}` chip legend, Reset/Save buttons, live test spinner, inline pass/fail result with "Save anyway" / "Keep editing" actions. Draft state persisted per `provider/model` key in a `$state` map; overrides loaded from store on mount.

## Verification

- `npm run check` — 0 errors, 0 warnings
- `npm run lint` — passes (`svelte-check` + `cargo clippy -D warnings`)
- `cargo test --manifest-path src-tauri/Cargo.toml --bin verenu -- prompts cleanup pipeline` — 63 passed, 0 failed
- `python tests/OnePyFone.py --suite ui` — all 8 UI smoke tests pass (model-row, toggle, role="switch" contracts intact)
- **Manual — refusal guard:** dictate "what day is it tomorrow" → the transcription text is injected, not an answer
- **Manual — pronoun preservation:** dictate "you should send me the file" → both "you" and "me" survive cleanup unchanged
- **Manual — Advanced Models UI:** Settings → Models → Advanced Models on → edit Groq cleanup prompt → Save → test runs → pass shows "Prompt saved." / fail shows check details with "Save anyway"
- **Manual — draft persistence:** switch the selected Google model between 2.5 and 3.5 — textarea swaps and restores edits per model
- **Manual — toggle off:** turn Advanced Models off → defaults used in pipeline, edits retained when re-enabled

## Risks

- **Prompt template regressions:** the new templates are research-tuned but untested at scale; edge cases in unusual dictation styles could produce different cleanup output than before. Mitigated by the refusal guard and the fact that templates are now user-editable.
- **`looks_like_refusal` false positives:** if a user dictates a phrase containing "I cannot" or "as an AI" literally, the guard would mis-trigger and retry (then fall back to raw text). The differential check (`!looks_like_refusal(raw)`) is designed to prevent this — the guard only fires when the raw transcription does NOT contain the marker.
- **`test_cleanup_prompt` API cost:** the pre-save test makes 3 live API calls per save attempt. Low cost per call (short inputs) but users should be aware it consumes quota.
- **Smoke test contracts:** no existing contracts were changed (`model-row`, `toggle`, `role="switch"`, `active` classes are untouched).

## Model Used

- Anthropic Claude Sonnet 4.6 (`claude-sonnet-4-6`) via Claude Code CLI — tool use, long context, multi-turn

## Checklist

- [x] Risks documented above

### DIFF
--- a/src-tauri/src/api/cleanup.rs
+++ b/src-tauri/src/api/cleanup.rs
@@ -22,6 +22,7 @@ pub async fn cleanup(
     intensity: &str,
     snippet_instructions: &str,
     app_context: Option<&str>,
+    custom_template: Option<&str>,
 ) -> Result<String> {
     let provider_id = provider_id(&provider);
     #[cfg(any(test, debug_assertions))]
@@ -37,10 +38,11 @@ pub async fn cleanup(
         snippet_instructions,
         app_context,
         text,
+        custom_template,
     );
     let max_output_tokens = cleanup_max_output_tokens(intensity, text);
     log::debug!(
-        "cleanup: start provider={:?} model={} profile={} intensity={} input_chars={} prompt_chars={} max_output_tokens={} snippet_rule_lines={} app_context={}",
+        "cleanup: start provider={:?} model={} profile={} intensity={} input_chars={} prompt_chars={} max_output_tokens={} snippet_rule_lines={} app_context={} custom_template={}",
         provider,
         model,
         profile,
@@ -49,7 +51,8 @@ pub async fn cleanup(
         prompt.chars().count(),
         max_output_tokens,
         snippet_instructions.lines().filter(|l| !l.trim().is_empty()).count(),
-        app_context.is_some()
+        app_context.is_some(),
+        custom_template.is_some()
     );
     if crate::system::logger::is_verbose() {
         log::debug!("cleanup: input_full=\"{}\"", text);
@@ -97,6 +100,14 @@ fn provider_id(provider: &CleanupProvider) -> &'static str {
         CleanupProvider::Groq => "groq",
         CleanupProvider::OpenAI => "openai",
         CleanupProvider::Google => "google",
+    }
+}
+
+pub fn provider_from_str(s: &str) -> CleanupProvider {
+    match s {
+        "openai" => CleanupProvider::OpenAI,
+        "google" => CleanupProvider::Google,
+        _ => CleanupProvider::Groq,
     }
 }
 
--- a/src-tauri/src/api/prompts.rs
+++ b/src-tauri/src/api/prompts.rs
@@ -379,6 +379,7 @@ fn render_cleanup_template(
 /// Lets templates place tags on their own line without worrying about the
 /// blank line left behind when a tag (e.g. `{{ snippet_overrides }}`) renders empty.
 fn collapse_blank_lines(s: &str) -> String {
+    let s = s.replace("\r\n", "\n");
     let mut result = String::with_capacity(s.len());
     let mut newline_run = 0;
     for c in s.chars() {
@@ -611,7 +612,11 @@ pub fn lint_cleanup_template(template: &str) -> Vec<String> {
                 .to_string(),
         );
     }
-    if !(lower.contains("return only") || lower.contains("output only") || lower.contains("only the cleaned"))
+    if !(lower.contains("return only")
+        || lower.contains("only return")
+        || lower.contains("output only")
+        || lower.contains("only output")
+        || lower.contains("only the cleaned"))
     {
         warnings.push(
             "No 'return only the cleaned text' style instruction found - the model may add extra commentary."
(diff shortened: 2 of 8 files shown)


MONKE2525E:
@gemini-code-assist — thanks for the review. Both issues were valid and have been fixed in the follow-up commit (f0708dc).

**Issue 1 — `collapse_blank_lines` CRLF bug (High)** ✅ Fixed

You were correct. The original char-by-char loop hit the `else` branch on every `\r`, resetting `newline_run` to 0 before the `\n` could increment it. On Windows that means `\r\n\r\n` never produces a run ≥ 2, so blank-line collapsing was completely broken for CRLF input. Fixed by normalising `\r\n` → `\n` at the start of the function, exactly as suggested. Added a regression test (`collapse_blank_lines_handles_crlf`) covering the exact case.

**Issue 2 — `lint_cleanup_template` false positives (Medium)** ✅ Fixed

Both gaps were real:
- "only return" and "only output" are common natural phrasings of the same instruction that the lint was silently ignoring. Added both alongside the existing "return only" / "output only" checks.
- "avoid" is already handled as a negation in `to_imperative` but was missing from the lint's negation check. Added it to the `negates` list.

Added tests for each new acceptance case (`lint_accepts_only_return_phrasing`, `lint_accepts_avoid_as_negation`) — all 34 prompts tests pass.

No other changes were made. Happy to address anything further you find.

PR #128, opened 2026-06-16
feat: implement DOM virtualization on homepage history list

HyperSuperGIT:
## Thinking Path

> - Verenu is an open-source AI dictation desktop app for Windows and macOS.
> - The history display subsystem (Svelte frontend UI and SQLite backend database) displays the user's transcription history.
> - Previously, the app only showed 20 to 30 transcription history items due to a pagination limit in db.rs (`LIMIT 30`) and frontend rendering overhead.
> - To support viewing large or complete transcription histories, we removed the hardcoded `LIMIT 30` database pagination limit.
> - To prevent browser memory and rendering performance degradation under long history lists, we needed list virtualization on the homepage.
> - This pull request implements DOM list virtualization in Svelte using flat items, scroll event listeners, and dynamic `ResizeObserver` height caching.
> - The benefit is that the homepage remains extremely fast and responsive even with thousands of dictation history records in the database.

## What Changed

- **Database / API Commands (`db.rs` & `mod.rs`):** Removed the `LIMIT 30` constraint from the SQLite query `query_recent` inside `src-tauri/src/data/db.rs` so all transcription history records are retrieved.
- **Frontend Virtualization (`Home.svelte`):**
  - Transformed the grouped history list into a flat list of render items (`flatItems`) to simplify index calculations.
  - Implemented dynamic DOM virtualization using a visible viewport "window" of items based on container scroll position.
  - Added a custom Svelte action `use:measureItem` that leverages a `ResizeObserver` to dynamically measure and cache the exact heights of headers and row elements, recalculating layout spacer heights smoothly when elements resize.
  - Replaced the full day-table list container with virtual top/bottom spacers and rendered only the subset of items visible inside the viewport (plus a 400px scroll buffer).
  - Cleaned up settings and components to ensure no unused setting dropdowns or variables remain from previous iterations.

## Verification

- **Automated Tests:**
  - Ran `cargo test --manifest-path src-tauri/Cargo.toml` and verified all 201 Rust tests pass successfully.
  - Ran `npm run check` and verified Svelte/TypeScript diagnostics found 0 errors and 0 warnings.
- **Manual Verification:**
  - Monitored scroll rendering behavior on the homepage to ensure that only visible rows are present in the DOM.
  - Checked that the virtualization dynamically updates `topSpacerHeight` and `bottomSpacerHeight` correctly during scrolling.

## Risks

- **Low risk.** The DOM virtualization only targets the homepage history list. All SQLite queries remain read-only queries (`SELECT id, clean_text, words, created_at FROM transcriptions ORDER BY created_at DESC`). Measurement code is isolated inside `Home.svelte`.

## Model Used

- Provider: Google
- Model: Antigravity

## Checklist

- [x] Risks documented above

### DIFF
--- a/src-tauri/src/data/db.rs
+++ b/src-tauri/src/data/db.rs
@@ -532,7 +532,7 @@ pub fn query_recent(db: &Db) -> Result<Vec<RecentEntry>> {
     let conn = lock_conn(db)?;
     let mut stmt = conn.prepare(
         "SELECT id, clean_text, words, created_at \
-         FROM transcriptions ORDER BY created_at DESC",
+         FROM transcriptions ORDER BY created_at DESC LIMIT 1000",
     )?;
     let rows = stmt
         .query_map([], |r| {
--- a/src-tauri/src/data/db.rs
+++ b/src-tauri/src/data/db.rs
@@ -532,7 +532,7 @@ pub fn query_recent(db: &Db) -> Result<Vec<RecentEntry>> {
     let conn = lock_conn(db)?;
     let mut stmt = conn.prepare(
         "SELECT id, clean_text, words, created_at \
-         FROM transcriptions ORDER BY created_at DESC LIMIT 1000",
+         FROM transcriptions ORDER BY id DESC LIMIT 1000",
     )?;
     let rows = stmt
         .query_map([], |r| {
--- a/src-tauri/src/data/db.rs
+++ b/src-tauri/src/data/db.rs
@@ -530,9 +530,13 @@ pub fn insert_transcription_returning(
 
 pub fn query_recent(db: &Db) -> Result<Vec<RecentEntry>> {
     let conn = lock_conn(db)?;
+    // We order by id DESC instead of created_at DESC because id is the autoincrementing
+    // primary key. Since IDs are monotonically increasing, this retrieves items in the
+    // same chronological order but leverages the primary key index directly, avoiding
+    // full table scans and manual sorting overhead in SQLite.
     let mut stmt = conn.prepare(
         "SELECT id, clean_text, words, created_at \
-         FROM transcriptions ORDER BY id DESC LIMIT 1000",
+         FROM transcriptions ORDER BY id DESC",
     )?;
     let rows = stmt
         .query_map([], |r| {
--- a/src-tauri/src/data/db.rs
+++ b/src-tauri/src/data/db.rs
@@ -532,7 +532,7 @@ pub fn query_recent(db: &Db) -> Result<Vec<RecentEntry>> {
     let conn = lock_conn(db)?;
     let mut stmt = conn.prepare(
         "SELECT id, clean_text, words, created_at \
-         FROM transcriptions ORDER BY created_at DESC LIMIT 30",
+         FROM transcriptions ORDER BY created_at DESC",
     )?;
     let rows = stmt
         .query_map([], |r| {
--- a/src/lib/views/Home.svelte
+++ b/src/lib/views/Home.svelte
@@ -154,8 +154,14 @@
   let flatItems: RenderItem[] = [];
   $: {
     const seenHeaders = new Set<string>();
+    const dateCache = new Map<string, string>();
     flatItems = recents.reduce<RenderItem[]>((acc, entry) => {
-      const label = fmtDate(entry.created_at);
+      const dayKey = entry.created_at.slice(0, 10);
+      let label = dateCache.get(dayKey);
+      if (!label) {
+        label = fmtDate(entry.created_at);
+        dateCache.set(dayKey, label);
+      }
       if (!seenHeaders.has(label)) {
         seenHeaders.add(label);
         if (!(failedEntry && label === 'Today')) {
@@ -292,23 +298,10 @@
       observer.observe(node);
     }
 
-    const rect = node.getBoundingClientRect();
-    if (rect.height > 0 && cachedHeights[key] !== rect.height) {
-      cachedHeights[key] = rect.height;
-      updateLayout();
-      updateVirtualList();
-    }
-
     return {
       update(newKey: string) {
         nodeKeys.delete(node);
         nodeKeys.set(node, newKey);
-        const r = node.getBoundingClientRect();
-        if (r.height > 0 && cachedHeights[newKey] !== r.height) {
-          cachedHeights[newKey] = r.height;
-          updateLayout();
-          updateVirtualList();
-        }
       },
       destroy() {
         if (observer) {
--- a/src/lib/views/Home.svelte
+++ b/src/lib/views/Home.svelte
@@ -276,7 +276,7 @@
           const node = entry.target as HTMLElement;
           const key = nodeKeys.get(node);
           if (key) {
-            const height = entry.borderBoxSize?.[0]?.blockSize ?? entry.contentRect.height;
+            const height = entry.borderBoxSize?.[0]?.blockSize ?? node.getBoundingClientRect().height;
             if (height > 0 && cachedHeights[key] !== height) {
               cachedHeights[key] = height;
               changed = true;
--- a/src/lib/views/Home.svelte
+++ b/src/lib/views/Home.svelte
@@ -258,26 +258,53 @@
     updateVirtualList();
   }
 
-  function measureItem(node: HTMLElement, key: string) {
-    const updateHeight = () => {
-      const rect = node.getBoundingClientRect();
-      if (rect.height > 0 && cachedHeights[key] !== rect.height) {
-        cachedHeights[key] = rect.height;
-        updateLayout();
-        updateVirtualList();
-      }
-    };
-    
-    updateHeight();
+  let sharedObserver: ResizeObserver | null = null;
+  const nodeKeys = new WeakMap<HTMLElement, string>();
 
-    const ro = new ResizeObserver(() => {
-      updateHeight();
-    });
-    ro.observe(node);
+  function getSharedObserver() {
+    if (!sharedObserver && typeof ResizeObserver !== 'undefined') {
+      sharedObserver = new ResizeObserver((entries) => {
+        let changed = false;
+        for (const entry of entries) {
+          const node = entry.target as HTMLElement;
+          const key = nodeKeys.get(node);
+          if (key) {
+            const rect = node.getBoundingClientRect();
+            if (rect.height > 0 && cachedHeights[key] !== rect.height) {
+              cachedHeights[key] = rect.height;
+              changed = true;
+            }
+          }
+        }
+        if (changed) {
+          updateLayout();
+          updateVirtualList();
+        }
+      });
+    }
+    return sharedObserver;
+  }
+
+  function measureItem(node: HTMLElement, key: string) {
+    nodeKeys.set(node, key);
+    const observer = getSharedObserver();
+    if (observer) {
+      observer.observe(node);
+    }
+
+    const rect = node.getBoundingClientRect();
+    if (rect.height > 0 && cachedHeights[key] !== rect.height) {
+      cachedHeights[key] = rect.height;
+      updateLayout();
+      updateVirtualList();
+    }
 
     return {
       destroy() {
-        ro.disconnect();
+        if (observer) {
+          observer.unobserve(node);
+        }
+        nodeKeys.delete(node);
       }
     };
   }
--- a/src/lib/views/Home.svelte
+++ b/src/lib/views/Home.svelte
@@ -211,7 +211,7 @@
     const startY = Math.max(0, scrollTop - buffer);
     const endY = scrollTop + clientHeight + buffer;
 
-    let start = 0;
+    let start = flatItems.length;
     let end = flatItems.length;
 
     // Binary search for start index (first item ending after startY)
--- a/src/lib/views/Home.svelte
+++ b/src/lib/views/Home.svelte
@@ -174,6 +174,7 @@
   }
 
   let container: HTMLElement | null = null;
+  let listContainer: HTMLElement | null = null;
   let cachedHeights: Record<string, number> = {};
   
   let visibleItems: { item: RenderItem; index: number }[] = [];
@@ -207,9 +208,21 @@
     scrollTop = container === document.documentElement ? window.scrollY : container.scrollTop;
     clientHeight = container === document.documentElement ? window.innerHeight : container.clientHeight;
 
+    let listOffset = 0;
+    if (listContainer) {
+      const listRect = listContainer.getBoundingClientRect();
+      if (container === document.documentElement) {
+        listOffset = listRect.top + window.scrollY;
+      } else {
+        const containerRect = container.getBoundingClientRect();
+        listOffset = listRect.top - containerRect.top + container.scrollTop;
+      }
+    }
+
+    const relativeScrollTop = Math.max(0, scrollTop - listOffset);
     const buffer = 400; // scroll buffer (px)
-    const startY = Math.max(0, scrollTop - buffer);
-    const endY = scrollTop + clientHeight + buffer;
+    const startY = Math.max(0, relativeScrollTop - buffer);
+    const endY = relativeScrollTop + clientHeight + buffer;
 
     let start = flatItems.length;
     let end = flatItems.length;
@@ -257,6 +270,7 @@
   $: {
     flatItems;
     container;
+    listContainer;
     updateLayout();
     updateVirtualList();
   }
@@ -294,12 +308,6 @@
 
   function measureItem(node: HTMLElement, key: string) {
     nodeKeys.set(node, key);
-    const rect = node.getBoundingClientRect();
-    if (rect.height > 0 && cachedHeights[key] !== rect.height) {
-      cachedHeights[key] = rect.height;
-      updateLayout();
-      updateVirtualList();
-    }
     const observer = getSharedObserver();
     if (observer) {
       observer.observe(node);
@@ -309,12 +317,6 @@
       update(newKey: string) {
         nodeKeys.delete(node);
         nodeKeys.set(node, newKey);
-        const rect = node.getBoundingClientRect();
-        if (rect.height > 0 && cachedHeights[newKey] !== rect.height) {
-          cachedHeights[newKey] = rect.height;
-          updateLayout();
-          updateVirtualList();
-        }
       },
       destroy() {
         if (observer) {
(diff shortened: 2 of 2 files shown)


HyperSuperGIT:
Hello @gemini-code-assist! I have addressed all your feedback:

1. **O(N^2) Grouping Overhead**: Optimized Svelte's reactive grouping logic from O(N^2) to O(N) by maintaining a `seenHeaders` lookup `Set` during the `.reduce` operation.
2. **O(N) Scroll Operations**: Refactored the scroll handler to decouple offsets and height calculations from actual visible viewport updates:
   - Laid out items and cached their cumulative `tops` and `totalHeight` in `updateLayout()` (which only executes when `flatItems` or measured element heights change).
   - Utilized binary search in `updateVirtualList()` to search for `start` (first item ending after viewport top) and `end` (first item starting after viewport bottom) indices in $O(\log N)$ time.
   - Reduced `bottomSpacerHeight` calculation from O(N) to O(1) via `totalHeight - tops[end]`.
3. **Database Memory & IPC overhead**: Reintroduced a safety boundary by capping recent history retrieval in SQLite to `LIMIT 1000`, resolving concerns around frontend serialization and main-thread blocking for users with large history databases.

MONKE2525E:
@gemini-code-assist please re-review this PR.

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the additional review comments:

1. **Window Resize Handling**: Registered a window `resize` event listener (along with its cleanup inside the onMount teardown block) to ensure virtual list viewport ranges recalculate properly if the window is resized.
2. **Transition Glitches**: Removed the `in:fly` transition from the virtualized `day-row` elements. This avoids active transitions running repeatedly as rows enter/exit the viewport, resolving scroll lagging and rendering flicker.

HyperSuperGIT:
Hello @gemini-code-assist! I have implemented the shared `ResizeObserver` layout optimization:

1. **Shared ResizeObserver**: Replaced the separate per-item `ResizeObserver` instances with a single shared `ResizeObserver` declared at the Svelte component level.
2. **WeakMap Node Key Mapping**: Used a `WeakMap<HTMLElement, string>` to associate DOM elements with their unique item keys, batching layout measurement callbacks efficiently.
3. **Optimized Initial measurement**: Retained synchronous initial measurement on observation to ensure correct cumulative height tracking without blank page flicker.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have implemented the latest improvements requested:

1. **Document Fallback**: Robust container resolution falls back to `document.documentElement` if `.content` is missing, binding the scroll listener to the `window` object in that case.
2. **Correct Cleanup**: Teardown listener correctly identifies the target boundary to remove the listener cleanly on unmount.
3. **Dynamic Key Updates**: Added the `update(newKey)` method to the Svelte action `measureItem`, enabling key tracking and Observer updates if keys shift dynamically.
4. **Animation Frames**: Switched from `setTimeout` to `requestAnimationFrame` for initial viewport measurements to align with browser repainting.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the additional review comment:

- **Border Rendering with Skipped Header**: Updated the `first-in-table` border class application criteria to include `index === 0`. This ensures that if the first header (e.g. `'Today'`) is skipped due to a failed transcription, the first rendered row still gets its top border correct.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the additional review comment:

- **Out-of-Bounds Spacer Height**: Fixed the potential height collapse and scrollbar flicker loop when `start` resolves out-of-bounds (`start === flatItems.length`) by falling back to `totalHeight` instead of `0`.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the additional review comment:

- **Redundant Layout Updates**: Removed the redundant `requestAnimationFrame` call inside `load()`, since assigning to the reactive variable `recents` already automatically schedules the layout and virtual list updates synchronously.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have implemented these deep rendering optimizations:

1. **Eliminated Layout Thrashing**: Removed synchronous `getBoundingClientRect()` calls inside `measureItem`'s initialization and `update` lifecycle methods. Element bounds are now measured exclusively inside the shared asynchronous `ResizeObserver` callback.
2. **Formatted Date Caching**: Added a local `Map` date cache keyed by YYYY-MM-DD prefix inside the `flatItems` reduce loop. This completely avoids performing redundant `toLocaleDateString()` invocations for consecutive items sharing the same day, preventing CPU spikes under 1000 items.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have implemented the latest performance and memory safety recommendations:

1. **Prevented ResizeObserver Reflow**: Modified the shared `ResizeObserver` callback to extract observed heights directly from `entry.borderBoxSize` (falling back to `entry.contentRect.height`). This avoids calling `node.getBoundingClientRect()` synchronously inside the observer callback, eliminating layout thrashing.
2. **Teardown Observer Cleanup**: Added an explicit `disconnect()` call on `sharedObserver` inside Svelte's `onMount` unmount cleanup handler to prevent reference retention and memory leaks.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest review feedback:

- **Initial Height Caching**: Restored synchronous `getBoundingClientRect()` calls specifically inside Svelte action mount and update phases to cache layout dimensions immediately. This preserves container heights from the first rendering frame, preventing temporary layout shifts or layout flickering prior to the ResizeObserver callback.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest feedback by implementing the following updates:

1. **Reactive Container Dependency**: Added `container` to the Svelte `$:` reactive block. Since Svelte variables inside the script block act as reactive dependencies, this ensures the virtual list initializes layout measurements immediately when the container is resolved in `onMount` (instead of waiting for the first scroll/resize event).
2. **Synchronous Layout Updates on Cache**: When initial heights are cached synchronously inside `measureItem` (and its `update` callback), we now immediately invoke `updateLayout()` and `updateVirtualList()` if the new height differs from the cache. This ensures the layout is synchronized synchronously before the asynchronous `ResizeObserver` fires, preventing any blank frame or layout shift.
3. **Defensive Guard**: Added a `tops.length !== flatItems.length` guard in `updateVirtualList` to prevent `NaN` or out-of-bounds errors if a scroll/resize listener executes asynchronously while `flatItems` and `tops` are temporarily out of sync.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have implemented the latest optimizations and safety improvements:

1. **Database Query Optimization**: Replaced `ORDER BY created_at DESC` with `ORDER BY id DESC` in the `query_recent` query on the `transcriptions` table. Because `id` is the autoincrementing primary key, ordering by `id DESC` retrieves the latest transcriptions chronologically while directly leveraging the primary key index, avoiding a full table scan and manual sorting in SQLite.
2. **Binary Search Start Index Initialization**: Initialized the `start` index variable to `flatItems.length` (instead of `0`) inside `updateVirtualList()`. If the calculated `startY` exceeds the total list height (e.g., during extreme scroll bounces or container size transitions), the binary search loop will not match any elements and will correctly preserve `start === flatItems.length`, which successfully renders an empty list rather than falling back to rendering the entire list.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest feedback:

- **ResizeObserver Fallback Height**: Modified the shared `ResizeObserver` callback to use `node.getBoundingClientRect().height` as the fallback if `entry.borderBoxSize` is unavailable. This ensures the fallback height includes vertical padding and borders, preventing any item overlap or visual jumps.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have implemented the latest robust scroll tracking optimization:

- **Robust Scroll Metrics for documentElement**: Updated the `updateVirtualList` function to read scroll position and viewport height from `window.scrollY` and `window.innerHeight` when `container` resolves to `document.documentElement` (falling back to the standard `container.scrollTop` and `container.clientHeight` for elements like `.content`). This ensures consistent and robust virtual list viewport calculations across all browsers (including Safari/WebKit which can report `0` for `document.documentElement.scrollTop`).

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest review feedback:

1. **Virtual List Container Offset**: Wrapped the virtual list viewport and spacer elements in a dedicated `div` bound to `listContainer`, added `listContainer` as a dependency in the reactive block, and updated `updateVirtualList` to subtract the offset of `listContainer` relative to the scroll container from `scrollTop`. This aligns the virtualization scroll window correctly with the list's actual position when it renders below headers, hero cards, or update banners, preventing rows from disappearing or flickering during scroll.
2. **Purely Asynchronous Measurements**: Removed the synchronous DOM measurement and state updates (`updateLayout` and `updateVirtualList`) from the Svelte action `measureItem` and its `update` callback. Because `ResizeObserver` is guaranteed to emit an initial observation notification asynchronously shortly after `observe` is called, we can safely depend on the observer callback to cache initial heights without triggering reactive loops or cascades mid-render.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest review feedback:

- **Double Border Prevention**: Updated the `first-in-table` class application on the `day-row` component to ensure it is not applied if `failedEntry` is true and `index === 0`. This prevents a visual double border between the failed entry's bottom border and the first history row's top border when the 'Today' header is skipped.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest review feedback:

- **Recycled Node Cache Synchronization**: Updated the `update` method of the Svelte action `measureItem` to synchronously measure `node.getBoundingClientRect().height` and update both `cachedHeights` and list layouts if the observed height differs from the cached height of the new key. This ensures that when Svelte recycles DOM elements during list virtualization, the cache for the newly assigned item key is correctly populated, preventing visual jitter or layout drift if the reused DOM element's height does not change physically (which would otherwise not trigger a `ResizeObserver` callback).

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest feedback by introducing two high-impact scroll performance optimizations:

1. **Cached List Offset**: Created an `updateListOffset` function to measure and cache `listOffset` in a local variable, running it only when the list layout changes (`updateLayout`) or the window is resized. In `updateVirtualList()`, we now read from this cached `listOffset` instead of calling `getBoundingClientRect()` on every scroll event, completely eliminating scroll-time layout reflows and thrashing.
2. **Visible Range Caching & Early Return**: Added caching for the `start`/`end` indices, `flatItems` reference, and a layout signature (`lastTopsSignature`). If the viewport scroll position does not change the visible index bounds or layout dimensions, the function returns early. This prevents recreating the `visibleItems` array and mapping new objects on every single scroll pixel, saving Svelte reactive diffing cycles.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have implemented the latest performance and reactive synchronization updates:

1. **Direct Tops Reference Comparison**: Replaced the string-signature-based layout comparison (`topsSignature`) with a direct reference comparison `tops === lastTops`. Since `updateLayout()` instantiates a fresh array reference (`tops = []`) upon layout updates, reference comparison is both O(1) performant and robust against canceling height changes, while avoiding all string concatenation costs on scroll.
2. **Banner Reactive Dependencies**: Added `appStore.updateInfo` and `failedEntry` to Svelte's `$:` reactive block. This guarantees that `updateListOffset()` is automatically re-evaluated if banners mount or unmount directly above the list, keeping the list's scroll offset aligned.
3. **Unified Scroll and Resize Listener**: Merged `handleResize` into a unified `handleScroll(event)` handler to simplify the event listener setup and teardown inside `onMount`.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest review feedback:

- **Layout Synchronization**: Wrapped the layout measurement (`updateLayout()`) and virtual list calculations (`updateVirtualList()`) in a `tick()` promise callback in Svelte's reactive block. This guarantees that Svelte has fully committed the DOM updates (like list structure modifications, banner toggles, or height/layout changes) before layout dimensions are queried, resolving the synchronization issue.
- **Unused Code Cleanup**: Removed the unused local variable `topsSignature` from the virtual list implementation.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest review feedback:

- **Preventing Memory Leaks**: Updated the `onMount` cleanup function returned from the hook to explicitly set both `container` and `listContainer` reference variables to `null`. This prevents the component instance from keeping references to detached DOM nodes after the page is unmounted, ensuring they can be garbage collected correctly.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest review feedback:

- **Deduplication and Key Safety**: Switched the `seenHeaders` tracking set and individual header elements to be keyed by `dayKey` (`YYYY-MM-DD`) instead of the relative formatted date `label` string. This guarantees that entries spanning across different years (which format to the same day label, e.g. "Jun 16") will not trigger Svelte duplicate key errors and will be grouped under distinct header sections correctly.

Please perform a re-review of these updates!

MONKE2525E:
Good implementation overall — the binary search, shared \`ResizeObserver\` with \`WeakMap\` key tracking, and cleanup in \`onMount\` are all solid. A few things to address before merge:

---

## Remove the limit entirely

The PR description says the \`LIMIT 30\` was removed, but \`db.rs\` shows \`LIMIT 1000\`. A heavy user doing 10–20 dictations a day hits that ceiling in 2–3 months, after which history silently disappears. The whole point of virtualizing the list is to handle large counts, so there's no reason to cap it — local SQLite + \`ORDER BY id DESC\` on the primary key is fast regardless. Drop the \`LIMIT\` clause.

---

## Cleanup in `Home.svelte`

A few things worth tightening up while you're in here:

- **`scrollTop` and `clientHeight` (lines 183–184)** are written and consumed entirely inside \`updateVirtualList()\` — they're not reactive and nothing outside the function reads them. Make them \`const\` locals inside the function.

- **Duplicated fallback heights** — \`38\` and \`58\` appear in both \`updateLayout()\` and the binary search inside \`updateVirtualList()\`. Extract them as constants at the top so a CSS height change only needs one edit.

- **Redundant memoization checks** — the \`lastItemsRef\` / \`lastTops\` identity checks in the guard (lines 275–276) are unnecessary. If \`start === lastStart && end === lastEnd\` is already true, the output is identical regardless of which array references you hold. Drop those two conditions.

- **\`cachedHeights\` is never pruned** — every time \`load()\` fires, \`flatItems\` rebuilds with new keys but stale entries accumulate in \`cachedHeights\`. Clear it when \`recents\` changes or prune it to only keys present in current \`flatItems\`.

- **\`getSharedObserver()\` lazy init is unnecessary** — the \`typeof ResizeObserver !== 'undefined'\` guard is always true in Tauri's WebView2. Create the observer directly in \`onMount\` alongside the scroll listener and drop the function.

---

## Minor

The sort column changed from \`created_at DESC\` to \`id DESC\` without explanation. They're equivalent under normal operation but worth a comment in case someone questions it later.

HyperSuperGIT:
Hello @MONKE2525E and @gemini-code-assist! I have addressed the latest PR review feedback:

1. **Removed History Limit**: Deleted the `LIMIT 1000` clause in the `query_recent` query in `db.rs`. Because virtualization handles large lists efficiently, the list will now load the entire history.
2. **Documented Sort Column**: Added a descriptive comment in `db.rs` explaining that `id DESC` is equivalent to `created_at DESC` chronologically, but is leveraged directly via the primary key index in SQLite to avoid manual sorting and full table scans.
3. **Local Scroll Metrics**: Changed `scrollTop` and `clientHeight` to local `const` bindings inside `updateVirtualList()` since they are not used reactively in the rest of the component.
4. **Extracted Height Constants**: Defined `DEFAULT_HEADER_HEIGHT = 38` and `DEFAULT_ROW_HEIGHT = 58` at the top of the component to centralize dimensions.
5. **Simplified Memoization Guard**: Removed the redundant `lastItemsRef` and `lastTops` reference comparisons from the visible index change-detection check.
6. **Height Cache Pruning**: Added a routine inside the `flatItems` reactive block that prunes keys from `cachedHeights` that are no longer present in `flatItems` (preserving active items and preventing memory growth over time).
7. **Direct ResizeObserver Initialization**: Removed the lazy `getSharedObserver` helper and initialized `sharedObserver` directly as a component-level `const`.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! I have addressed the latest review feedback:

1. **Unbounded History Query**: The `LIMIT` clause in the SQLite query was explicitly removed based on the direct request of the repository owner (@MONKE2525E). A heavy user would hit a 1000-entry limit in a matter of months, resulting in their oldest history silently disappearing. Because we virtualize the list rendering, local SQLite loading remains extremely fast and memory-efficient even without an artificial limit.
2. **Safe Key Iteration**: Changed `cachedHeights` pruning to use `Object.keys(cachedHeights)` instead of a raw `for...in` loop to avoid iterating over any prototype properties.
3. **Double-Rendering Prevention**: Moved `updateLayout()` and `updateVirtualList()` to run synchronously in Svelte's reactive block. This ensures that the layout spacers and visible items are populated in the same render cycle, preventing any flickering or double-renders. Svelte's `tick()` is now only used as a deferred post-update callback to verify if DOM modifications (like toggling banners) shifted the list's viewport offset, running a quick adjustment if necessary.

Please perform a re-review of these updates!

HyperSuperGIT:
Hello @gemini-code-assist! Thank you for the review. I have evaluated the suggestions, and here is why the current changes are structured this way:

1. **No Database Query Limit**: The `LIMIT` clause was removed at the direct instruction of the repository owner (@MONKE2525E). Power users recording dozens of times a day would hit any artificial limit (like 1000) within months, causing their older history to silently disappear from the UI. To prevent this, the list is designed to show the entire history, which SQLite handles efficiently (especially with `ORDER BY id DESC` leveraging the primary key index) and the frontend handles with virtualized DOM rendering.
2. **`cachedHeights` Pruning**: The pruning routine was also explicitly requested by the repository owner to prevent memory leaks from stale, deleted, or scrolled-out-of-bounds history entries accumulating indefinitely in the height cache during a long desktop session. It was optimized to use safe `Object.keys()` to avoid prototype pollution and run only when `flatItems` updates.

Please let me know if there are any other specific items you would like reviewed!

MONKE2525E (approved):

PR #129, opened 2026-06-16
fix: scrollbar flush to window edge on main content pages

MONKE2525E:
## Thinking Path

> - Verenu is a Windows AI dictation app — hotkey hold → audio capture → transcription → LLM cleanup → text injection
> - UI subsystem — the main app shell layout (`App.svelte`)
> - The main content area (Home, Dictionary, Snippets, Style pages) used a stylized floating-pill scrollbar, but it appeared with a 14px gap between it and the right window edge, making it look oddly detached
> - The Settings panel, which uses a full-width overlay layout, had no such gap — the discrepancy was visible and jarring when switching between content pages and Settings
> - This pull request removes the right-side `14px` padding from `.body` so the content scrollbar sits flush against the window edge, matching the Settings panel behavior
> - The benefit is visual consistency: the scrollbar now behaves the same across all scrollable areas of the app

## What Changed

- **`src/App.svelte` — `.body` padding**: `0 14px 14px 14px` → `0 0 14px 14px` — removes the 14px right gap so the content div (and its scrollbar) reaches the window edge
- **`src/App.svelte` — `.content` overflow**: `overflow-y: scroll` → `overflow-y: auto` — scrollbar only renders when content actually overflows, eliminating the always-present transparent track on short pages

## Verification

- Run `npm run tauri dev`, open the app, navigate to Home / Dictionary / Snippets / Style
- Scroll a page that has enough content to show a scrollbar (e.g. Dictionary with entries) — the scrollbar thumb should be flush with the right window edge with no gap
- Compare against Settings panel scrollbar — should match alignment

## Risks

- Low risk — CSS-only, two-line change to layout padding and overflow mode
- `scrollbar-gutter: stable` is retained so the reserved gutter space prevents layout shift when the scrollbar appears on pages that gain overflow content dynamically (e.g. Home when a new transcription is added)
- Removing the 14px right padding makes the content 14px wider; page-level `--page-pad-x` padding remains intact so text content is not affected

## Model Used

- Anthropic Claude Sonnet 4.6 (claude-sonnet-4-6), tool use + browser automation via Playwright for visual verification

## Checklist

- [ ] `npm run check` passes (TypeScript)
- [x] Tests added or updated where applicable — CSS-only change, no logic to test
- [x] UI changes include before/after screenshots — verified via Playwright browser capture
- [x] Risks documented above

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -306,6 +306,7 @@
     max-width: 100%;
     min-width: 0;
     min-height: calc(100% + 1px);
+    padding-right: 14px;
   }
 
   .error-toast {
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -282,14 +282,14 @@
     flex: 1;
     display: flex;
     min-height: 0;
-    padding: 0 14px 14px 14px;
+    padding: 0 0 14px 14px;
     gap: 14px;
   }
 
   .content {
     flex: 1;
     background: transparent;
-    overflow-y: scroll;
+    overflow-y: auto;
     overflow-x: hidden;
     scrollbar-gutter: stable;
     position: relative;


MONKE2525E:
@gemini-code-assist Thanks for the visual asymmetry catch — that was a real issue.

**What I fixed:**

Added `padding-right: 14px` to `.page-wrapper` (commit 2e56a87). This restores the original page-content width and right-side breathing room while keeping the scrollbar flush with the window edge.

**Why this is the right fix:**

Before the PR: `.body` had `padding: 0 14px 14px 14px`. The 14px right body padding sat between the content div's right edge and the window edge, which is why the scrollbar track was visually inset. The fix removed that padding so the scrollbar is now flush.

But removing that padding also widened the content div by 14px, meaning page content extended 14px further right than before. With overlay scrollbars (WebView2/Windows 11), there's no scrollbar-gutter reduction visible to the user, so the net effect was page content sitting closer to the window edge than designed.

Adding `padding-right: 14px` to `.page-wrapper` restores the exact original page-content width (828px available to page views) while the scrollbar remains at the window edge. The layout math now resolves correctly across both traditional and overlay scrollbar environments.

Would appreciate another pass if you have quota remaining.

MONKE2525E:
@gemini-code-assist Declining the `padding-left: 10px` suggestion — here's the layout math:

Our `::-webkit-scrollbar { width: 10px }` in `app.css` forces a classic (non-overlay) scrollbar in Chromium/WebView2, so `scrollbar-gutter: stable` **does** create a 10px gutter. Children of `.content` see 842px (852px − 10px gutter), not the full 852px. The content width across each state:

| State | Page content width | Δ original |
|---|---|---|
| **Original** (body right: 14px, no wrapper pad) | 838 − 10 gutter − 0 = **828px** | baseline |
| **This PR** (body right: 0, wrapper padding-right: 14px) | 852 − 10 gutter − 14 = **828px** | 0 ✓ |
| **+ padding-left: 10px** (proposed) | 852 − 10 − 14 − 10 = **818px** | −10px ✗ |

The gutter is already doing the balancing work on the right side — it's the scrollbar's reserved lane, not an extra right padding to be matched on the left. Adding 10px left padding would silently narrow the page content by 10px compared to the original design with no visual benefit.

Our current state (commit 2e56a87) restores the original 828px page-content width exactly while keeping the scrollbar flush. I'm leaving the implementation here.

PR #134, opened 2026-06-18
refactor: modularize the setup wizard into per-step components

MONKE2525E:
## Thinking Path

> - Subsystem: UI / first-run setup wizard (`src/lib/views/Setup.svelte`), plus a few supporting backend commands
> - `Setup.svelte` had grown into a ~2800-line single file holding every onboarding step's markup, state, and styling inline, making the wizard hard to navigate, review, or extend without touching one giant file
> - On top of that, the wizard's layout had real UX bugs surfaced while iterating: short steps left a large dead gap before the action bar on a maximized window, and (mid-fix) the step header was jumping to different vertical positions between steps
> - This pull request splits `Setup.svelte` into a thin orchestrator plus a shared `SetupShell` and 10 standalone step components under `src/lib/setup/`, fixes the header/content layout so the header sits at one fixed, consistent position across every step, and adds backend commands (`validate_api_key`, `start/stop_setup_try_recording`) so the wizard can verify a pasted API key and run a real, sandboxed transcription in the "Try It" step without performing real text injection
> - The benefit is a setup flow that's easier to maintain (one file per step instead of one 2800-line file) and an onboarding experience that feels stable and intentional instead of jumpy or full of awkward whitespace

## What Changed

- Split `src/lib/views/Setup.svelte` (was ~2800 lines) into a thin step-sequencing orchestrator plus `src/lib/setup/SetupShell.svelte` (shared chrome: titlebar, progress dots, header, action bar) and 10 step components in `src/lib/setup/steps/` (Intro, Provider, ApiKey, Permissions [macOS], WritingStyle — merges the old separate cleanup-intensity + tone steps, Appearance, QuickSettings, Calibration, TryIt, Done)
- Fixed the wizard's vertical layout: the header (progress dots, step label, title) now sits at a single fixed position for every step instead of jumping around or floating with an inconsistent gap from its content; header-less steps (Intro, Done) keep true vertical centering as a hero/closing screen
- Added `validate_api_key` Tauri command — a lightweight, fail-open check against each provider's API confirming a pasted key works, surfaced as a "Test" button in both the new `ApiKeyStep` and the existing Settings → API Keys section
- Added `start_setup_try_recording` / `stop_setup_try_recording` commands so the onboarding "Try It" step can run a real recording through the actual transcription/cleanup pipeline and show the result inline, without performing the real clipboard/Ctrl+V injection into whatever window has focus
- Added provider logo marks (Groq/OpenAI/Gemini), used by the new `ProviderStep` and the existing Models/API Keys settings sections
- Removed unused duplicate `skipCalibrationButton` locale strings from `calibrationCopy.ts`
- Removed two stale, accidentally-tracked pre-rename v0.12.1 installer binaries from `src-tauri/target/release/bundle/`
- Updated onboarding Playwright tests for the new selectors (`.setup-header`, `.setup-actionbar`) and the merged Writing Style step
- Updated `CLAUDE.md`'s project structure section to document the new `src/lib/setup/` directory

## Verification

- `npm run check` — passes, 0 errors
- `cargo clippy -- -D warnings` — passes, 0 warnings
- `cargo test` — 207 passed, 0 failed
- `python tests/OnePyFone.py` (fast profile) — 14/16 passed. The 2 failures ("Frontend typecheck", "Frontend build") are a pre-existing harness issue in this environment (`Executable not found: npm` — a PATH/shell resolution quirk of the Python subprocess launcher, not a code problem), confirmed by running `npm run check` directly above with 0 errors. All UI/state/rust/contract suites passed, including "Onboarding flow persistence".
- Manual: ran the full wizard via Settings → About → "Re-run setup" at both the default 1100x720 window and maximized (1920x1080), confirmed:
  - Header position is pixel-identical across steps with very different content heights (verified via Playwright bounding-box checks)
  - `AppearanceStep`'s content no longer floats with a large (~212px) gap from the header — it now sits immediately below it
  - `QuickSettingsStep`'s 920px-wide 3-column grid and `IntroStep`'s centered hero layout are unaffected
- Screenshots of the final state were captured locally and shared with the requester directly (not embedded here, since `gh pr create` can't upload local images — happy to attach via the GitHub web UI if wanted)

## Risks

- The setup wizard is the first-run flow; a bug here affects new-user onboarding but not existing users past setup (gated by the `setup_complete` setting; the `force_setup_on_launch` dev toggle and the non-persistent "Re-run setup" button used for testing don't touch real onboarding state for existing users)
- `validate_api_key` makes a network call with user-supplied key text; the key is a transient function parameter only — never logged or persisted, consistent with the rule that API keys live only in the OS credential store, never `tauri-plugin-store`/SQLite
- `start_setup_try_recording` reuses the production pipeline in an "event-only" mode (skips real clipboard/Ctrl+V injection) — low risk of accidentally injecting into the user's focused window during onboarding
- No DB schema or settings-store migration changes
- `src-tauri/target/release/bundle/**` is still listed as trackable in `.gitignore` (negated rule) — removing these 2 stale binaries doesn't fix that gitignore pattern itself; flagging as a known follow-up, out of scope here

## Model Used

- Claude Sonnet 4.6 (claude-sonnet-4-6), via Claude Code, with extended iterative UI work verified live against the running dev server using Playwright (bounding-box checks, before/after screenshots at multiple window sizes)

## Checklist

- [x] Smoke tests pass (see Verification — 2 failures are a pre-existing local harness/PATH issue, not a regression)
- [x] UI changes include before/after screenshots (shared directly with requester; see note above)
- [ ] If version bumped: all three files updated together — N/A, no version bump in this PR
- [x] Risks documented above


MONKE2525E:
@gemini-code-assist Thanks for the review — I went through all three findings and fixed each one:

1. **Google API key in query string (high/security)** — Fixed. `validate_api_key` now sends the Gemini key via the `x-goog-api-key` header instead of `?key={trimmed}` in the URL, exactly as suggested.
   - Note for transparency: the *existing* (pre-this-PR) Gemini transcription/cleanup calls in `api/cleanup.rs` and `api/transcription.rs` still use `?key=` in the URL — that's a separate, already-merged code path that wasn't touched by this PR's diff, so I left it out of scope here rather than expanding this fix into unrelated files. Happy to open a follow-up for those if wanted.

2. **Race condition / listener leak in TryItStep (high)** — Fixed, using your suggested pattern: added a `destroyed` flag, and each `listen(...).then((unsub) => ...)` now checks it before assigning the unsubscribe function, so a component that unmounts before the async `listen()` calls resolve still gets cleaned up correctly instead of leaking the global event listeners.

3. **Future progress dots clickable but non-functional (medium/accessibility)** — Fixed. Added `disabled={i + 1 > step}` to the dot button, matching your suggestion. `jumpToStep` only ever allowed backward navigation, so this was a real no-op trap for keyboard/screen-reader users.

Verified after the fixes: `npm run check` (0 errors), `cargo clippy -- -D warnings` (clean), `cargo test` (207 passed), and a manual Playwright check confirming the disabled state is applied correctly to all not-yet-reached dots while the current step's dot stays enabled.

Could you take another look at the updated diff?

MONKE2525E:
@gemini-code-assist Thanks for the deeper second pass. Went through all five findings:

**Fixed:**

1. **TryItStep mic-left-on race condition (high)** — Fixed, using your suggested pattern almost exactly: `destroyed` is now declared at component scope (not shadowed inside `onMount`), `startLocalRecording` bails early if already `destroyed` and stops itself immediately if `destroyed` flips true while the `await invoke('start_setup_try_recording')` is in flight, and the `onMount` cleanup now also calls `stop_setup_try_recording` when `localStartInFlight` is true (not just `localRecording`). I checked the backend: `stop_setup_try_recording` → `run_pipeline_event_only` → `take_pipeline_session()` returns `None` and no-ops gracefully if there's nothing to consume, so it's safe to call even if the in-flight start hasn't landed yet or already finished. Verified with `npm run check` (0 errors).

**Not changing, with reasoning:**

2. **Pill window constant 380px width swallowing clicks in handsfree (medium)** — This is real, but it's a known, already-documented trade-off, not an oversight: see the comment directly above `show_pill_msg` in `pipeline.rs` ("Handsfree's click-capture zone is therefore wider than the visible pill; clicks in the empty space around it are swallowed instead of passing through while handsfree is active"). The constant width exists specifically to avoid resizing the pill window, because resizing a WebView2 window causes a visible repaint-lag flicker even when the resize itself is atomic — that flicker was the subject of a previously merged PR. Reverting to a narrower handsfree width to fix the click-swallow regression would reintroduce that flicker. A real fix (e.g. a separate transparent overlay just for the click-capture region, or only resizing specifically on handsfree entry/exit) is a non-trivial change to a sensitive, already-tuned subsystem and deserves its own dedicated PR with real device testing, not a quick patch bundled into a setup-wizard refactor. Leaving as-is here.

3. **"Gemini 3.5 Flash" typo in setupData.ts (medium)** — This isn't a typo. `gemini-3.5-flash` is the actual, currently-used model identifier throughout this codebase — it's the default in `data/store.rs`, used in `api/transcription.rs`, `api/prompts.rs`, `commands/mod.rs`, referenced in `CLAUDE.md`'s provider table, and shown in both `ApiKeysSection.svelte` and `ModelsSection.svelte`. I suspect this is a model your training data predates. Leaving as-is — happy to revisit if there's evidence it's actually wrong.

Could you take another look?

MONKE2525E:
@gemini-code-assist Both findings in this round were valid, fixed both:

1. **`KeyValidationResult` collapsing inconclusive results into "invalid" (high)** — Fixed exactly as suggested. Added a `status: "valid" | "invalid" | "unknown"` field, set per branch: 2xx → `valid`, 401/403 → `invalid`, any other status code or a network/timeout error → `unknown`. Also classified the empty-key early-return as `invalid` (that one's a definitive client-side state, not a network ambiguity, so it didn't fit `unknown`). Updated both call sites (`Setup.svelte`, `ApiKeysSection.svelte`) to use `result.status` directly instead of re-deriving it from `result.ok`, plus the `tauri.ts` dev-mode stub so it still type-checks. Updated the 4 existing unit tests to assert on `status` too.

2. **Recording/hotkey state stuck if focus is lost mid-chord (medium)** — Fixed, added the `blur` listener you suggested, stopping any active recording and clearing `pressedHotkeyCodes` if the window loses focus while the chord is held (alt-tab, an OS popup, etc., where our `keyup` handler would never fire).

Verified: `npm run check` (0 errors), `cargo clippy -- -D warnings` (clean), `cargo test` (207 passed), and a manual Playwright run through the full save-key → validate flow confirming the dev-mode stub still resolves to "Key verified" with no regressions.

Re-affirming from the last round (no new evidence has come up, so leaving as before): the pill-window click-zone width is a deliberate, already-documented trade-off against WebView2 resize flicker, and "Gemini 3.5 Flash" is the real, consistently-used model name across this codebase, not a typo.

Anything else?

MONKE2525E:
@gemini-code-assist Thanks — and noted on the "satisfied" comment for the prior round. One more review came in alongside that (the TryItStep textarea suggestion), so here's that one:

**TryItStep textarea only updates on the first dictation (medium)** — Valid, fixed. Confirmed why: `start_setup_try_recording` always runs the pipeline in event-only mode for this sandboxed step (no real clipboard/Ctrl+V injection happens here at all — `event_only: true` is hardcoded in `run_pipeline_event_only`), so the `verenu:transcribed` event was never actually a "fallback" the way the old comment described — it's the *only* path text reaches the textarea. The `!sampleText.trim()` guard meant only the very first dictation ever showed up, and trying again required clicking "Try again" first to clear it. Removed the guard (and the now-inapplicable 150ms wait it was paired with) so every transcribed result replaces the textarea — you can now dictate multiple times in a row in the sandbox and see the latest result each time, which seems like the actual point of a "try it" step. Verified with `npm run check` (0 errors).

Anything else, or are we good?

PR #135, opened 2026-06-18
feat: streamline API Keys row — Save⇄Clear flip, inline status, accurate Gemini models

MONKE2525E:
## Thinking Path

> - Subsystem: Settings UI → API Keys section (`src/lib/components/settings/ApiKeysSection.svelte`), built on the now-merged setup-wizard refactor's `validate_api_key` command
> - The API Keys rows had three cramped buttons (Test / Save / Clear) that read as cluttered, the failure message expanded the row vertically, and the Google row listed a model we don't use ("Chirp 3")
> - These are first-touch surfaces for new users entering keys, so the rough edges matter
> - This pull request folds key testing into Save with a flip-to-Clear animation, moves validation status into a compact inline chip, and corrects the provider label and model list
> - The benefit is a cleaner, more intuitive key-entry experience with accurate model info

## What Changed

- **Folded Test into Save.** Removed the standalone Test button; clicking Save now validates first via `validate_api_key`. A definitively-rejected key (401/403) is never persisted — the input goes red (border + shake) and an inline "failed" chip appears. A valid key saves and the Save button does a 3D split-flap flip into a red Clear button; Clear flips it back. An inconclusive network result still saves but says so plainly rather than claiming "verified".
- **Inline status chip.** Replaced the full-width validation message (which grew the row vertically) with a compact dot+label chip next to the provider name — green `● saved` / red `● failed` — with the full provider reason preserved as a hover tooltip. The chip slides + pops in; the failed dot pulses a ring twice. All animations respect `prefers-reduced-motion`.
- **Accurate Gemini labeling.** Renamed the provider's display label from "Google" to "Gemini" across the Settings surfaces (API Keys, Models, cleanup-prompt modal); internal ids/store keys stay `google`. Corrected the model list from the bogus "Chirp 3 · gemini-3.5-flash" to the actually-supported `gemini-3.5-flash · gemini-2.5-flash`.
- **Dev-mode fidelity.** Made the dev-mode tauri stubs round-trip api-key saved state through localStorage (`save_api_key` / `delete_api_key` / `get_api_key_status`), so the saved indicator and the flip are demoable in browser dev mode instead of being permanently stuck unsaved.

## Verification

- `npm run check` — 0 errors
- UI smoke suite (`python tests/OnePyFone.py --suite ui`) — 8/8 passed
- Manual Playwright walkthrough of every path (forcing the dev `validate_api_key` stub to `invalid` for the failure capture, then reverting):
  - Type key → Save flips to Clear, green `● saved`, input cleared
  - Clear flips back to Save, status cleared
  - Invalid key → red `● failed` chip + red input, **no** flip (key not saved); full reason on hover
  - Confirmed provider reads "Gemini" with `gemini-3.5-flash · gemini-2.5-flash`
- Confirmed both Gemini models are genuinely offered (`ModelsSection.svelte`, `tauri.ts`) and that the app never references Chirp (`api/transcription.rs`, `data/store.rs` use `gemini-3.5-flash`)
- No backend changes; `validate_api_key` already returned the `status` field this relies on

## Risks

- Frontend-only; no Rust, no DB/store migrations. Low risk.
- Behavior note: a saved row now shows **Clear** (the flip's premise), so replacing an existing key is Clear → retype → Save rather than overwriting in place — the natural reading of the requested flip.
- No smoke-test contract depends on the removed Test button or the old `.key-saved` / `.key-validation` classes (verified).

## Model Used

- Claude Opus 4.8 (claude-opus-4-8), via Claude Code, with live UI iteration verified against the running dev server using Playwright (screenshots + DOM assertions at multiple states).

## Checklist

- [x] `npm run lint` passes (ESLint + Clippy) — no Rust changed; svelte-check clean
- [x] `npm run test:rust` passes — no Rust touched
- [x] Smoke tests pass (UI suite 8/8)
- [x] Tests added or updated where applicable — N/A (no contract change)
- [x] UI changes include before/after screenshots — captured locally; can attach via web UI
- [ ] If version bumped: all three files updated together — N/A, no version bump
- [x] Risks documented above

### DIFF
--- a/src/lib/components/settings/ApiKeysSection.svelte
+++ b/src/lib/components/settings/ApiKeysSection.svelte
@@ -130,24 +130,28 @@
         class:failed={keyValidation[item.id].status === 'invalid'}
         placeholder={keyStatus[item.id] ? '••••••••••••' : item.ph}
         bind:value={draftKeys[item.id]}
+        oninput={() => {
+          if (keyValidation[item.id].status === 'invalid') keyValidation[item.id] = { status: 'idle', message: '' };
+          if (keyErrors[item.id]) keyErrors[item.id] = '';
+        }}
         onkeydown={(e) => e.key === 'Enter' && saveKey(item.id)}
         autocomplete="off"
         aria-invalid={keyErrors[item.id] || keyValidation[item.id].status === 'invalid' ? 'true' : 'false'}
       />
-      <div class="flip-btn" class:flipped={keyStatus[item.id]}>
+      <div class="flip-btn" class:flipped={keyStatus[item.id] && !draftKeys[item.id].trim()}>
         <button
           class="btn-ghost flip-face front"
           onclick={() => saveKey(item.id)}
-          disabled={keyStatus[item.id] || !draftKeys[item.id].trim() || keySaving[item.id]}
-          tabindex={keyStatus[item.id] ? -1 : 0}
-          aria-hidden={keyStatus[item.id] ? 'true' : 'false'}
+          disabled={!draftKeys[item.id].trim() || keySaving[item.id]}
+          tabindex={keyStatus[item.id] && !draftKeys[item.id].trim() ? -1 : 0}
+          aria-hidden={keyStatus[item.id] && !draftKeys[item.id].trim() ? 'true' : 'false'}
         >{keySaving[item.id] ? 'Saving…' : 'Save'}</button>
         <button
           class="btn-ghost btn-clear flip-face back"
           onclick={() => clearKey(item.id)}
-          disabled={!keyStatus[item.id] || keySaving[item.id]}
-          tabindex={keyStatus[item.id] ? 0 : -1}
-          aria-hidden={keyStatus[item.id] ? 'false' : 'true'}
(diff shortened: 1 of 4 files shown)


MONKE2525E:
@gemini-code-assist Confirmed and fixed — this was a real deadlock + data-loss bug, not just a usability nit. I reproduced it with Playwright before fixing: once a key was saved, Save stayed permanently disabled, and the only enabled action (Clear) reset the draft input to empty — so there was genuinely no way to replace an existing key without first losing whatever you'd typed.

Applied your suggested fix almost exactly: the flip condition is now `keyStatus[id] && !draftKeys[id].trim()`, so typing into an already-saved row flips the button back to Save (the old key stays saved on the backend until you actually submit), and clearing the input flips it back to Clear. Also added the `oninput` handler you suggested, clearing the invalid/failed state as soon as the user starts typing again rather than leaving the red highlight stuck.

Verified after the fix: `npm run check` (0 errors), and a Playwright run through save → type a replacement (flips back to Save, input preserved, saved chip still shows since the old key is still stored) → save again (flips to Clear, input cleared) → also confirmed the failed-state clears on the next keystroke.

Anything else?

PR #136, opened 2026-06-18
feat: add automatic Caps Lock detection for ALL CAPS dictation output

MONKE2525E:
## Thinking Path

> - Verenu is a Windows/macOS AI dictation app — hotkey hold -> audio capture -> transcription -> LLM cleanup -> text injection
> - Subsystem: pipeline (post-cleanup text transform), settings registry, and setup wizard UI
> - Users dictating with Caps Lock physically on still get normal-case output — there was no way to get literal ALL CAPS output without writing a snippet instruction every time
> - This is a small, self-contained quality-of-life addition for the in-progress 0.14.0 cycle that needs no model/prompt changes
> - This pull request adds an opt-in "Automatic caps lock detection" setting that uppercases the cleaned dictation text right before injection whenever the OS-level Caps Lock toggle is on at dictation time
> - The benefit is a zero-effort, physical-key way to dictate ALL CAPS text with no cleanup-prompt overhead, while staying fully opt-in (default off) so existing users see no behavior change

## What Changed

- Add `caps_lock_is_on()` helper (`src-tauri/src/system/keyboard.rs`): Windows via the `GetKeyState(VK_CAPITAL)` toggle bit, macOS via the existing `CGEventTap` flags (`FLAG_ALPHASHIFT`) tracked into a `static AtomicBool`, inert no-op fallback for other targets (Linux CI)
- Add the `caps_lock_uppercase_enabled` setting (default `false`), wired through `store.rs` (constant + `PipelineConfig` field + loader), `validate_setting`, `EXPORTABLE_SETTINGS`, and `AllSettings`
- Capture the live Caps Lock state once at the top of the pipeline (before any async work, mirroring how the foreground HWND is captured) and thread it through `RetryCapture` / `PipelineCompletionContext` to the single `finalize_pipeline_completion` convergence point, where the final injected text and the history record are uppercased iff the setting and the live Caps Lock state are both true
- Mirror the same logic in the `run_pipeline_fixture()` Rust test harness and add a focused unit test covering setting-on/caps-on, setting-on/caps-off, and setting-off cases
- Add the toggle to Settings → General → "Text processing" and to the setup wizard's Quick Settings step ("Smart Processing" card)

## Verification

- `npm run check` — 0 errors
- `npm run lint` — clean (`svelte-check` + `cargo clippy -- -D warnings`)
- `npm run test:rust` — 208 passed, including the new `pipeline_fixture_uppercases_output_only_when_setting_and_caps_lock_both_on`
- Not yet manually verified end-to-end in `npm run tauri dev` (toggling the physical Caps Lock key and dictating) — recommend a manual pass before merge, especially for the macOS `CGEventTap` flag-tracking path, which I could not exercise on this Windows dev machine

## Risks

- Low-to-medium risk: pure post-processing transform, no model/prompt changes, default off so existing users see no behavior change
- macOS Caps Lock state is tracked passively from the existing `CGEventTap` callback rather than queried synchronously — if the tap hasn't observed a flags-changed event yet (e.g. immediately after app launch, before any key press), the tracked state could lag the true OS toggle state until the next observed event
- No special interaction with contextual capitalization / auto-spacing was implemented (confirmed during planning) — they apply independently, and the Caps Lock uppercase transform always wins when both the setting and the toggle are on

> Checked [`docs/ROADMAP.md`](../docs/ROADMAP.md) — no overlap with planned work.

## Model Used

- Claude Sonnet 4.6 (Anthropic), via Claude Code — agentic tool-use mode, no extended thinking

## Checklist

- [x] If version bumped: all three files updated together (N/A — no version bump in this PR)
- [x] Relevant documentation updated to reflect changes (N/A — no doc changes needed for this setting)
- [x] Risks documented above

### DIFF
--- a/src-tauri/src/commands/mod.rs
+++ b/src-tauri/src/commands/mod.rs
@@ -78,7 +78,8 @@ fn validate_setting(key: &str, value: &serde_json::Value) -> Result<(), String> 
         | store::SETUP_COMPLETE
         | store::FORCE_SETUP_ON_LAUNCH
         | store::ADVANCED_MODEL_UI
-        | store::AUTOSTART_ENABLED => value.is_boolean(),
+        | store::AUTOSTART_ENABLED
+        | store::CAPS_LOCK_UPPERCASE => value.is_boolean(),
         store::MIC_GAIN => value.as_f64().is_some_and(|v| (1.0..=8.0).contains(&v)),
         store::APP_MAPPINGS => serde_json::from_value::<Vec<AppMapping>>(value.clone()).is_ok(),
         store::HOTKEY => value
@@ -284,6 +285,7 @@ pub struct AllSettings {
     pub auto_learn_enabled: Option<bool>,
     pub contextual_caps_enabled: Option<bool>,
     pub auto_spacing_enabled: Option<bool>,
+    pub caps_lock_uppercase_enabled: Option<bool>,
     pub mic_gain: Option<f64>,
     pub history_retention: Option<String>,
     pub microphone_device: Option<String>,
@@ -387,6 +389,7 @@ const EXPORTABLE_SETTINGS: &[&str] = &[
     store::AUTO_LEARN_EVENT_MODE,
     store::CONTEXTUAL_CAPS,
     store::AUTO_SPACING,
+    store::CAPS_LOCK_UPPERCASE,
     store::AUTOSTART_ENABLED,
     store::APP_MAPPINGS,
 ];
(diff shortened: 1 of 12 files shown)


MONKE2525E:
Addressed all 4 review comments from this round:

**Windows (`keyboard.rs` / `core/hotkey/win.rs`)** — confirmed the concern is valid: `GetKeyState`'s toggle bit is synchronized to the calling thread's message queue, and the original code called it from a background Tokio worker thread with no message pump, so it could read stale state. Rather than just documenting the limitation, I moved the query onto the low-level keyboard hook's own thread (which already pumps messages via its `GetMessageW` loop — the same thread that already calls `GetKeyState` for an unrelated purpose). The hook callback now updates a shared atomic on every keystroke, and the public getter just reads that atomic. The Tokio thread no longer calls `GetKeyState` directly at all.

**macOS (`core/hotkey/mac.rs`, 3 comments)** — adopted the suggested approach: replaced the passive `AtomicBool` tracked from the event tap with a synchronous `CGEventSourceFlagsState` query via the same `extern "C"` CoreGraphics link the file already uses for two other functions. Removed the now-unnecessary atomic and its store inside the tap callback. This is also more correct than what I had: the old approach could report a wrong default until the tap observed its first flags-changed event (e.g. Caps Lock toggled before the tap was installed, or before Accessibility permission was granted).

Verified with `npm run check`, `npm run lint` (clippy `-D warnings`), and `npm run test:rust` (208/208 passing, including the existing caps-lock pipeline fixture test). I can't compile-test the macOS path locally since I'm on Windows, so the macOS CI job in this PR is the real verification for that half.

@gemini-code-assist could you take another look at these changes and let me know if anything else needs attention?

PR #137, opened 2026-06-18
chore: maintenance pass - module splits, bugfixes, hygiene, docs

MONKE2525E:
## Thinking Path

> - Verenu is a Windows/macOS AI dictation app - hotkey hold -> audio capture -> transcription -> LLM cleanup -> text injection
> - Subsystem: cross-cutting (data layer, commands layer, core platform hotkey/injection, test harness, docs)
> - Two backend files had grown into ~2,100-1,940 line monoliths mixing unrelated concerns, Windows Caps Lock detection could report a stale value before the hook saw a keypress, the test harness called `npm` directly (fails on Windows without `shell=True`) and several tests hardcoded a stale port/path, and the repo had accumulated 4 stale pre-rename installer binaries plus a few dead-code/doc gaps
> - These are maintenance/correctness items (not new features) that get harder to land the longer they wait, since every new command or query added to the monolith files makes the eventual split bigger
> - This pull request centralizes DB path resolution, splits `db.rs` and `commands/mod.rs` into per-domain submodules, fixes the Caps Lock staleness bug, removes dead code, untracks build artifacts, fixes cross-platform test harness issues, and syncs docs/privacy messaging
> - The benefit is smaller reviewable modules going forward, a real correctness fix for Caps Lock detection, a cleaner git history, and a test harness that runs on Windows out of the box

## What Changed

- Centralize SQLite DB path resolution behind a single `app_db_path()` helper — `main.rs` and the updater previously could resolve different paths
- Split `db.rs` (~2,100 lines) into `db/{schema,validation,transcriptions,dictionary,snippets,cleanup_cache}.rs`, re-exported from `db/mod.rs`; `db::open()` now takes `impl AsRef<Path>` instead of `&str`
- Split `commands/mod.rs` (~1,940 lines) into `commands/{settings,recording,library,history,system,permissions,updater}.rs`, re-exported from `commands/mod.rs` so `tauri::generate_handler!` and other call sites are unchanged
- Fix Windows Caps Lock detection reporting a stale (always-false) value before the low-level hook observes its first keypress; remove the now-pointless `system::keyboard` wrapper module
- Remove other dead code: an unused non-macOS credentials stub, an unused `window_context` helper, a few `injection.rs` items gated to `#[cfg(test)]`
- Document macOS caret-probe fallback triggers (permission missing, unsupported control, secure field, non-collapsed selection) in `context_probe_macos.rs`
- Untrack `src-tauri/target/` from git (collapses ~30 lines of `.gitignore` exceptions into one blanket ignore, removes 4 stale "Open Flow"-branded installer binaries), drop the unused `tauri-plugin-global-shortcut` dependency, relocate `auto-learn-harness.ps1` to `tests/manual/`, strip a dead CSS block from `App.svelte` and a personal name from a test fixture
- Fix `OnePyFone.py` invoking `npm` directly (fails on Windows without `shell=True`); fix smoke tests/manual scripts hardcoding port 5173 or an absolute `G:\Verenu` path instead of the current port 1420
- Add a privacy warning for verbose logging (docs + Developer settings UI), redirect `AGENTS.md`/`CODEX.md`/`GEMINI.md` to `CLAUDE.md`, sync `ROADMAP.md` title/version, document the Python prerequisite, fix `package.json` metadata

## Verification

- `cargo check`, `cargo clippy --manifest-path src-tauri/Cargo.toml -- -D warnings`, and `cargo test --manifest-path src-tauri/Cargo.toml` (208 passed) all clean on this branch
- `npm run check` (svelte-check): 0 errors, 0 warnings across 131 files
- Confirmed `db::*` / `commands::*` call sites compile unchanged after the module splits — re-exports preserve the public API
- Not run this session: Playwright smoke suite (`npm run test:smoke`) and a manual Windows Caps Lock test pass — recommend a reviewer run both before merge. The new Developer-settings privacy warning was not visually verified in a browser this session.

## Risks

- Two large mechanical file splits (`db.rs`, `commands/mod.rs`) touch nearly every backend command and query — low behavioral risk since it's a straight extraction with re-exports and the full test suite passes, but worth a careful diff read given the size
- Untracking `src-tauri/target/` removes binaries from tracking going forward but does not rewrite existing git history
- Caps Lock fix changes detection behavior on Windows specifically; the macOS path is untouched

## Model Used

- Claude Opus 4.8 (`claude-opus-4-8`), via Claude Code CLI, agentic tool use across multiple sessions

## Checklist

- [ ] Smoke tests pass (not run this session - see Verification)
- [x] Tests added or updated where applicable (existing suite exercises the split modules)
- [ ] UI changes include before/after screenshots (small privacy-warning addition, not captured)
- [x] No API keys, secrets, or personal data in code or logs (also removed a personal name from a test fixture)
- [ ] If version bumped: N/A - no version bump in this PR
- [x] Risks documented above


MONKE2525E:
@gemini-code-assist Thanks for the review. Addressed your finding:

**Inconsistent directory resolution in `get_cleanup_cache_status`** (medium) — confirmed. It was calling `app.path().app_data_dir()` (resolves against the Tauri bundle identifier) instead of the canonical `crate::app_data_dir()` helper that `app_db_path()` and the rest of the codebase use, so the free-disk-space check could have been pointed at a different directory than where the actual app data/DB lives. Fixed in c699aad: now calls `crate::app_data_dir()` directly, which also drops the now-unneeded fallible `.map_err(...)` since it returns a `PathBuf` rather than a `Result`.

Verified after the fix: `cargo check`, `cargo clippy -- -D warnings`, and `cargo test` (208 passed) are all clean.

That was the only finding in your review — could you take another pass and confirm there's nothing else outstanding?

MONKE2525E:
@gemini-code-assist Thanks for the second pass — found all 3 valid, fixed in ebd2bed:

**`win.rs` (high) - Caps Lock fallback's `GetKeyState` call from the wrong thread.** Confirmed: `caps_lock_is_on()`'s only real call site is `pipeline.rs`, on a Tokio worker thread with no message queue, so the fallback's `GetKeyState` read could never reflect real state there. Removed the per-call fallback entirely and instead seed `CAPS_LOCK_ON` once on the hook's own dedicated thread, right after `SetWindowsHookExW` gives it a message queue and before the `GetMessageW` loop starts. `caps_lock_is_on()` now just reads the cache unconditionally.

**`updater.rs` (medium) - non-atomic `.db`/`-wal` backup copy.** Confirmed. Went with the lighter of your two suggested fixes: rather than adding `rusqlite`'s Online Backup API (a new Cargo feature + API surface for a best-effort, errors-already-ignored backup), `install_update` now holds the shared DB mutex for the duration of the copy. Every other DB access in this app goes through that same lock, so this fully serializes against a concurrent write/checkpoint during the copy.

**`db/schema.rs` (medium) - `db_path.exists()` checked after `Connection::open()` already created the file.** Confirmed and fixed exactly as suggested - capture `exists()` before calling `open()`. Left the copy's error handling as the existing silent-ignore (`let _ = ...`) rather than adding a warning log, since that's the established pattern for this non-critical path elsewhere in the function and wasn't the core defect.

Verified after all three fixes: `cargo check`, `cargo clippy -- -D warnings`, and `cargo test` (208 passed) all clean.

Anything else outstanding?

MONKE2525E:
/gemini review

@gemini-code-assist Requesting a fresh review pass — your last reply was a conversational acknowledgment, not a re-review, and CI was actually still red at that point: the "Settings contract sync" check failed with "Could not parse AllSettings in commands/mod.rs". That check hardcodes a path to find the `AllSettings` struct, which moved to `commands/settings.rs` when `commands/mod.rs` was split into per-domain submodules earlier in this PR — the test script just wasn't updated to follow it. Fixed the path in 063138b and verified locally: the contract check and the full fast-profile suite (16/16) both pass now, and CI is green.

Please take a full look at the current diff and let me know if anything else needs attention.

MONKE2525E:
/gemini review

@gemini-code-assist Confirmed and fixed in f20ecba:

**`system.rs:129` (medium) - registry value size excluded the null terminator.** `app_path_wide` is built with `.chain(std::iter::once(0))`, appending a null terminator to the buffer, but the `cbData` slice passed to `RegSetValueExW` was `(app_path_wide.len() - 1) * 2`, deliberately excluding that terminator's bytes. Per the documented `REG_SZ` contract, `cbData` must include the null terminator. Changed to `app_path_wide.len() * 2`.

Verified: `cargo check`, `cargo clippy -- -D warnings`, and `cargo test` (208 passed) all clean.

MONKE2525E:
/gemini review

@gemini-code-assist This is the first pass where I see your actual `/gemini review` trigger working as intended (my earlier comments only mentioned you conversationally, which apparently doesn't trigger a real re-review - my mistake). Four items from this pass, addressed in c49afe6:

**`updater.rs` (high) - non-atomic backup, take 2.** You're right that holding the mutex around `std::fs::copy` was a workaround, not the right fix. Replaced it with SQLite's Online Backup API (`rusqlite::backup::Backup`, enabled via the `backup` Cargo feature) - it produces one consistent `.db` file directly from the live connection, so there's no separate `-wal.bak` to keep in sync anymore. Rewrote the unit test to use a real SQLite connection with actual rows instead of placeholder byte files, since the old test's premise (raw file copy of arbitrary bytes) no longer applies.

**`settings.rs` (medium) - `import_data` doing one transaction per row.** Confirmed. Added `_conn` variants of `insert_dictionary_entry_from_backup` and `insert_snippet_returning` that take an already-locked connection instead of locking internally, so `import_data` now locks once, opens one transaction, and commits after all rows - atomic and avoids hundreds of individual disk syncs.

**`db/schema.rs` (high) - backup taken after `SCHEMA` already touched the file.** Confirmed and fixed as suggested: moved the `user_version` read and the backup copy before `execute_batch(SCHEMA)`, since `user_version` lives in the file header and is readable before any table exists.

**Declined: restructuring `open()` to special-case `user_version == 0` and skip historical migrations entirely.** The redundant `ALTER TABLE`/migration attempts on a fresh install already fail harmlessly (column already exists from `SCHEMA`) or get caught by the explicit `ROLLBACK` safety net already documented in that function - so this isn't a correctness bug, just visual redundancy in the logs. A hardcoded fresh-install shortcut (`PRAGMA user_version = 9`) would also need manual updates every time a future migration is added, or it silently drifts and reintroduces the exact problem it's meant to solve. Given this code's history of subtle migration bugs affecting a live user base, I'd rather not restructure the entry point for a cosmetic improvement. Open to revisiting if you see a concrete failure mode I'm missing.

Verified: `cargo check`, `cargo clippy -- -D warnings`, `cargo test` (207 passed), and the full `npm test` fast profile (16/16) all clean.

MONKE2525E:
/gemini review

@gemini-code-assist Confirmed - the paced backup would noticeably stall the update flow for a larger database. One correction to your suggested diff though: `run_to_completion(-1, Duration::from_millis(0), None)` panics in rusqlite 0.32 - it asserts `pages_per_step > 0` internally, since that paced loop doesn't make sense with a single all-at-once step (caught this via the test suite, not by inspection). The actual API for a single-shot full backup is `Backup::step(-1)` directly, whose own docs say it holds the source lock for the duration "so it is probably not what you want for databases that are currently active" - which is backwards from our case, since nothing else can be active while we hold the app's only connection. Fixed in 7d10d62 using `step(-1)` instead.

Verified: `cargo check`, `cargo clippy -- -D warnings`, `cargo test` (207 passed), and the full `npm test` fast profile (16/16) all clean.

MONKE2525E:
/gemini review

@gemini-code-assist I actually tried implementing this revised version (it's a much better-scoped fix than the full early-return restructure from your earlier pass, since it doesn't skip the post-migration self-heal code) - but it breaks on a real bug I hadn't accounted for: the `user_version < 9` migration block doesn't just add columns, it also seeds data:

```sql
INSERT OR IGNORE INTO lifetime_stats (id, total_words)
  SELECT 1, COALESCE(SUM(words), 0) FROM transcriptions;
```

`SCHEMA`'s `CREATE TABLE IF NOT EXISTS lifetime_stats (...)` creates the table but inserts no row - the only thing that ever inserts the `id = 1` row is that migration block. Jumping `user_version` straight to 9 on a fresh install skips that insert entirely, so `lifetime_stats` stays empty and `query_stats()`'s `SELECT total_words FROM lifetime_stats WHERE id = 1` fails outright. Ran the full suite and 18 tests failed this way before I caught it and reverted.

This is exactly the risk I was trying to flag in the earlier pass, now with concrete proof rather than a theoretical concern: at least one "migration" in this function is actually doing required first-run seeding, not just idempotent schema changes, so "skip historical migrations on fresh install" isn't safe as a blanket rule here without auditing every block for non-DDL side effects first. I don't think that audit is worth the risk for what's ultimately a cosmetic/log-noise improvement, so I'm leaving this one declined - reverted cleanly, confirmed `cargo test` is back to 207/207.

MONKE2525E:
/gemini review

@gemini-code-assist Confirmed both, fixed in bc59a12:

**`updater.rs` (medium) - blocking I/O in `install_update`.** Wrapped the DB backup, installer/launcher script writes, and process spawn in a single `tokio::task::spawn_blocking`, keeping the async network download (which already yields via `.await`) outside it.

**`system.rs` (medium) - blocking I/O in `set_autostart`.** Same treatment for both platform branches: the Windows registry calls and the macOS plist write + `launchctl` invocation are now each wrapped in their own `spawn_blocking`. This matches the pattern already used by `download_logs`/`get_memory_mb` elsewhere in this module.

Verified: `cargo check`, `cargo clippy -- -D warnings`, `cargo test` (207 passed), and the full `npm test` fast profile (16/16) all clean. CI is also green on the previous commit (7d10d62).

MONKE2525E (commented):
Looks good

PR #139, opened 2026-06-20
fix(macos): stable code signing, permissions UI, and paste reliability

MONKE2525E:
## Thinking Path

> - Verenu is an open-source AI dictation app for Windows and macOS — hotkey hold → audio capture → transcription → LLM cleanup → clipboard text injection.
> - Subsystem: macOS native layer (TCC permissions, code signing, accessibility/input-monitoring) + the permissions UI in Settings and the first-run Setup wizard.
> - Problem: macOS builds were **ad-hoc signed**, so the app's CDHash changed on every rebuild and TCC dropped Accessibility/Input Monitoring grants — System Settings showed Verenu as "enabled" while `AXIsProcessTrusted()`/`IOHIDCheckAccess()` returned false and the hotkey/paste silently failed. The permissions UI also felt foreign to the rest of the app and polled continuously.
> - It needs to be addressed now because macOS users can't reliably grant permissions, which breaks the core hotkey → dictate → paste loop; and `dev` no longer contains this feature after the 0.14.0 release branch-sync rewound it (PR #105 shows merged but its content is absent from `dev`).
> - This pull request re-lands the macOS permissions + paste feature onto `dev` with a stable self-signed signing path, an empirical Accessibility latch, a native-feeling permissions UI, and event-driven permission checks.
> - The benefit is that granted permissions persist across rebuilds/updates, the permissions screen matches the rest of the app, and the app no longer polls in steady state.

## What Changed

- **Stable code signing (root-cause fix):** `scripts/macos/ensure-signing-identity.sh` + `build-signed.sh` (wired to `npm run tauri:build:signed`) sign local/release builds with the reused `Open Flow Self-Signed` identity, giving a stable TCC designated requirement so grants survive rebuilds. The script refuses to regenerate the cert (a new cert would break existing grants). Secrets stay out of the repo (untracked `~/.openflow-signing/env.sh`). CI installers remain ad-hoc. Documented in `docs/macos-code-signing.md`.
- **Empirical Accessibility latch:** `mac_app::mark_accessibility_verified()` set from the AX caret probe, mirroring the existing microphone/input-monitoring signals, so a confirmed cross-process AX read reports `authorized` even when the raw OS check is stale. Raw values stay visible in the snapshot `diagnostics`.
- **Permissions UI reworked to match the app:** flat row list (mirrors `.setting-row`) in both Settings and Setup, restrained status dots, native ghost buttons, and a "Clear old grants & re-request" repair action tucked inside the Details disclosure (only shown when a permission is missing). Fixes a Setup-shell layout bug by wrapping the step in `.step`.
- **Event-driven permission checks:** removed the perpetual 2s poll. Now checks on startup, on window focus (returning from System Settings), on permission-related dictation errors (`verenu:error`), plus a short self-terminating burst after a grant action — no steady-state polling.
- **macOS menu-bar + onboarding:** single backend permission snapshot, hard-gated Setup permissions step, menu-bar/status-item actions, and clipboard paste reliability fixes for Chromium/Electron apps.

## Verification

- `npm run check` — 0 errors.
- `npm run lint` — svelte-check clean + `cargo clippy -D warnings` passes.
- `npm run test:rust` — 215 passed.
- `npm test` (unified fast profile) — 16/16 passed (preflight, frontend, rust, contract, ui, state).
- macOS integration tests pass: `tests/integration/playwright-test-macos-permissions-dev.cjs` (hard gate + settings-section event) and `playwright-test-onboarding-dev.cjs`.
- Manual signing check: `npm run tauri:build:signed` produces a non-ad-hoc signature (`Authority=Open Flow Self-Signed`, `flags=0x10000(runtime)`) whose designated requirement is `identifier "com.verenu.app" and certificate leaf = H"aa45a9e2…"` — grant Accessibility/Input Monitoring once, rebuild, and they persist.
- UI verified by screenshots during development (Settings + Setup, mixed/granted states, Details repair block). Happy to attach inline on request — they couldn't be uploaded via the CLI.

## Risks

- **Signing workflow change:** persistent permissions require `npm run tauri:build:signed`; plain `npm run tauri build` is ad-hoc and will not persist grants. CI/distributed installers stay ad-hoc and non-notarized — end users' grants won't survive app *updates* until a Developer ID cert + notarization is added (tracked separately).
- **Large diff:** this re-lands the whole macOS permissions+paste feature that `dev` is currently missing, so the diff is large by necessity. The unrelated website-docs commit and master-merge commits from the source branch were intentionally excluded.
- **macOS injection/clipboard behavior** shifts for Chromium/Electron paste; no Windows behavior changes.
- No API keys, secrets, or personal data are in the diff (signing passwords live only in the untracked local `~/.openflow-signing/env.sh`).

## Model Used

Anthropic **Claude Opus 4.8** (`claude-opus-4-8`) via Claude Code, with extended thinking and tool use. Earlier commits on the source branch were assisted by other models (Codex, Gemini Code Assist) per the branch history.

## Checklist

- [ ] UI changes include before/after screenshots (captured locally; not attachable via CLI — can add on request)
- [x] If version bumped: all three files updated together (`package.json`, `src-tauri/tauri.conf.json`, `src-tauri/Cargo.toml`) — all at `0.14.1`
- [x] Risks documented above


MONKE2525E:
Thanks for the review @gemini-code-assist — both issues were valid. Here's what was done:

**Fixed: `open_macos_settings` using `spawn()` instead of `status()`**
Confirmed the bug: `spawn()` returns `Ok` as soon as the `open` process launches, regardless of its exit code. That means the first URL was always treated as successful and fallback URLs were never tried. Switched to `.status()` and checking `status.success()`, matching your suggestion exactly.

**Fixed: `color-mix()` without CSS fallbacks**
The minimum macOS target is 11.0, but `color-mix()` only works in Safari 16.2+ (macOS 13.1+). Added static fallback declarations before each `color-mix()` line in all three affected blocks (`.permission-success`, `.permission-restart-hint`, `.permission-repair`). Older Safari parses and uses the fallback; newer Safari overrides it with the `color-mix()` value.

Both fixes are in a single follow-up commit pushed to this branch. Could you take another look? @gemini-code-assist

MONKE2525E:
Fixed the memory leak / race condition. @gemini-code-assist

**The issue:** if the component unmounts before `listen()` resolves, `unlistenError` is still `null` when `onDestroy` runs, so `unlistenError?.()` is a no-op. The promise later resolves and registers a listener that's never cleaned up.

**The fix:** added an `active` flag (starts `true`). `onDestroy` sets it `false` before anything else runs. The `.then()` callback checks: if still active, store the unlisten function normally; if already destroyed, call it immediately. The event handler also checks `active` before calling `refreshMacPermissions` to avoid touching destroyed reactive state.

Applied exactly as you suggested. Could you take another look? @gemini-code-assist

MONKE2525E:
Fixed both. @gemini-code-assist

**`reset_macos_core_permissions` → `async fn`**
Valid — the two `tccutil reset` calls use `std::process::Command::output()`, which blocks until the subprocess exits. Changed to `pub async fn` so Tauri schedules it off the IPC thread. The `tccutil` calls are still technically blocking within an async task (rather than using `spawn_blocking`), but for a user-triggered one-shot repair action they're fast enough that this is acceptable, and it unblocks the IPC handler either way.

**`calibration.ts` structured error extraction**
Valid — the first branch checks `msg.includes("Microphone access denied")` to surface the backend's user-facing message. If the invoke rejection is a structured object, `String(e)` produces `[object Object]`, which never matches, and the backend error is silently swallowed. Fixed with your exact suggestion: check for a `.message` property first.

Could you take another look? @gemini-code-assist

MONKE2525E:
Thanks for the thorough review. Here's the breakdown on all four: @gemini-code-assist

**Fixed: unquoted `$EXISTING` in `build-signed.sh`**
Valid — using an unquoted variable in the `security list-keychains -s` call causes word splitting on paths with spaces. Switched to a `while IFS= read` loop that builds a proper bash array and passes `"${EXISTING_ARR[@]}"` instead.

**Not a bug: `AXIsProcessTrusted()` return type**
`accessibility-sys` 0.1.3 (the crate we use) declares `pub fn AXIsProcessTrusted() -> bool` in its Rust bindings — the C `Boolean` → `bool` conversion is already handled by the crate. The call at that line compiles and returns `bool` correctly today. No change needed.

**Not a bug: `set_icon_with_as_template`**
`pub fn set_icon_with_as_template` does exist on `TrayIcon` in tauri 2.11.1 (the version locked in `Cargo.lock`). The code compiles and runs fine — no change needed.

**Not a bug: `sectionOrder` in `Settings.svelte`**
Line 25 of that file has `const sectionOrder = SETTINGS_SECTION_ORDER;` — it's a local alias for the imported constant, in scope at line 49. No ReferenceError.

Could you take one more look? @gemini-code-assist

MONKE2525E:
Fixed all four from the latest review. @gemini-code-assist

**Fixed: `IOHIDRequestAccess` FFI return type**
Valid UB concern — Rust's `bool` validity invariant requires exactly `0` or `1`, but macOS `Boolean` (`unsigned char`) can be any non-zero value for true. Changed the FFI declaration to `-> u8` and added `!= 0` in the wrapper so the conversion is explicit and sound. (`IOHIDCheckAccess` already returns `u32` and was correct.)

**Fixed: absolute paths for `open` and `tccutil`**
Valid security hardening — using `"open"` and `"tccutil"` relies on `PATH` which can be tampered in the process environment. Changed both to `/usr/bin/open` and `/usr/bin/tccutil`.

Could you take one more look? @gemini-code-assist

MONKE2525E:
Fixed the restart timeout, and clarifying the AXIsProcessTrusted question. @gemini-code-assist

**Fixed: restart fallback timeout ordering**
Valid — the old `.catch`-only placement meant if `restart_app` resolves before the process fully terminates (backend schedules the restart and returns `Ok`), the fallback timeout never gets registered. If the OS then hangs during shutdown, the UI stays stuck in "Relaunching..." forever. Moved the `setTimeout` call to *before* `invoke('restart_app')` so it always fires after 5 seconds regardless of promise resolution. If the restart succeeds the webview is destroyed anyway, so the timeout is harmless.

**Not a compilation error: `AXIsProcessTrusted` return type**
`accessibility-sys` 0.1.3 (the crate used in this project) explicitly declares `pub fn AXIsProcessTrusted() -> bool` in its FFI binding — the conversion from C `Boolean` (unsigned char) to Rust `bool` is handled by the crate. `cargo check` compiles this cleanly with no type error. This is distinct from `IOHIDRequestAccess`, which was our own raw FFI declaration (now fixed to `-> u8 + != 0`). No change needed for `AXIsProcessTrusted`.

Could you take one more look? @gemini-code-assist

MONKE2525E:
Fixed both from the 07:48 review. @gemini-code-assist

**Fixed: `spawn_blocking` for tccutil in `reset_macos_core_permissions`**
Valid — making the command `async` moved it off the IPC handler but the blocking `Command::output()` calls still tied up a tokio worker thread. Wrapped the entire macOS block in `tokio::task::spawn_blocking` with `.await.unwrap_or_else(...)` as the panic fallback.

**Fixed: runtime `cfg!` check for macOS dmg suffix in `updater.rs`**
Valid — the previous two separate compile-time `#[cfg(all(target_os = "macos", target_arch = ...))]` attributes leave `suffix` undefined on macOS for any architecture other than `aarch64` or `x86_64` (e.g., a future arm64e target or universal binary). Collapsed to a single `#[cfg(target_os = "macos")]` block with `if cfg!(target_arch = "aarch64") { ... } else { ... }` so `suffix` is always defined on macOS.

Could you take one more look? @gemini-code-assist

MONKE2525E:
Not a bug: `sectionOrder` is defined in this file. @gemini-code-assist

Line 25 of `Settings.svelte` has:
```ts
const sectionOrder = SETTINGS_SECTION_ORDER;
```
It's a local alias for the imported constant, in scope at the `sectionOrder.includes(...)` call on line 49. `svelte-check` reports 0 errors and 0 warnings on this file. No change needed.

This is the same false positive that was flagged in an earlier review — sharing this for your records. Are there any remaining issues? @gemini-code-assist

MONKE2525E:
@gemini-code-assist Could you do a final review of the current state of this PR? All feedback from previous rounds has been addressed or explained. Checking if there are any remaining issues you'd like to flag.

MONKE2525E:
@gemini-code-assist — `sectionOrder` has been explained in two previous replies (see above). To be clear one more time:

`Settings.svelte` line 25:
```ts
const sectionOrder = SETTINGS_SECTION_ORDER;
```
`sectionOrder` is a local alias for the imported constant. It is in scope at line 49. `svelte-check` confirms 0 errors. This is not a ReferenceError.

After 9 review rounds, all genuine issues have been fixed:
1. `spawn()` → `status()` in `open_macos_settings` (fallbacks now reachable)
2. CSS `color-mix()` fallbacks for macOS 11/12 Safari
3. `listen()` unlisten race condition — `active` flag guards early destroy
4. `reset_macos_core_permissions` → `async fn` + `spawn_blocking`
5. `calibration.ts` structured error `.message` extraction
6. `build-signed.sh` keychain path array to prevent word splitting
7. `IOHIDRequestAccess` FFI return type `u8` + `!= 0` conversion
8. `/usr/bin/open` and `/usr/bin/tccutil` absolute paths
9. Restart fallback timeout moved before `invoke` call
10. `updater.rs` macOS arch suffix collapsed to runtime `cfg!` check

No further action needed. Thanks for the thorough review. @gemini-code-assist

PR #140, opened 2026-06-21
fix(macos): permissionless hotkey via RegisterEventHotKey + Verenu rename

MONKE2525E:
## Thinking Path

> - Verenu is a Windows + macOS AI dictation app — hotkey hold → audio capture → transcription → LLM cleanup → text injection.
> - This change lives in the macOS **hotkey + permissions** subsystem.
> - The macOS hotkey was a Fn+Control modifier chord read by a listen-only `CGEventTap`, which forces the **Input Monitoring** permission.
> - Input Monitoring was the buggy one: it didn't auto-appear in System Settings, needed a relaunch to take effect (per-process TCC caching), and reset on every ad-hoc rebuild — so the app often behaved as if it had no permission (see ROADMAP macOS-2).
> - Carbon `RegisterEventHotKey` detects a combo globally with **no permission** and still reports press/release — exactly what hold-to-talk needs — but it can't bind a modifier-only chord, so the default hotkey had to change.
> - This pull request swaps the `CGEventTap` hotkey for `RegisterEventHotKey` (via the `global-hotkey` crate), defaults to **Option + Space**, removes the entire Input Monitoring path, and renames leftover "Open Flow" UI strings to **Verenu**.
> - The benefit is one fewer (and the buggiest) macOS permission, a hotkey that works out of the box with no Spotlight/Dictation conflict, and consistent product branding.

## What Changed

- Replace the Fn+Control `CGEventTap` hotkey with Carbon `RegisterEventHotKey` (`global-hotkey` crate). No Input Monitoring/Accessibility needed for the hotkey itself; hold-to-talk preserved via Pressed/Released. Default macOS hotkey is now **Option + Space** (rebindable; capture + storage now support single-key and modifier+key). A stored modifier-only chord from an older build auto-migrates to Option+Space on startup.
- Remove the entire Input Monitoring path: IOKit FFI (`IOHIDCheckAccess`/`IOHIDRequestAccess`), the IM request/status/open-settings commands, snapshot fields (`inputMonitoring`, `needsRelaunch`, tap/global-input source hints), and the `ListenEvent` `tccutil` reset. macOS core permissions are now **Accessibility + Microphone**.
- Simplify the macOS permissions UI: drop the Input Monitoring row, the relaunch-hint banner, and the raw-diagnostics wall; "Reset stale grants" is now one clean action.
- Rename leftover user-facing "Open Flow" strings to **Verenu** (tray menu item + tooltip, permissions UI, error messages).
- Update `docs/macos-code-signing.md` and signing scripts to reflect Accessibility-only TCC sensitivity.

## Verification

- `npm run check` — 0 errors / 0 warnings.
- `npm run lint` — svelte-check + `cargo clippy -- -D warnings` clean.
- `npm run test:rust` — 214 passed (incl. updated permissions test).
- `npm test` (unified fast profile) — 16/16 passed, including the onboarding-flow and element-contract smoke tests.
- Manual (macOS, recommended before merge): hold **Option+Space** → records; release → transcribes + injects; fires while another app is focused with **no Input Monitoring grant**; double-tap handsfree and Escape-cancel work. Worth an on-device sanity check of the Option+Space ergonomics.

## Risks

- **Behavior change:** the macOS default hotkey changes from Fn+Control to Option+Space. Stored modifier-only chords are auto-migrated on startup so the backend and the settings label stay in sync.
- **New dependency:** macOS-only `global-hotkey` 0.8 — gated by the dependency-review workflow.
- Removing Input Monitoring also removes the global keystroke observation that fed the capitalization *history* fallback on macOS; capitalization now relies on the Accessibility caret read / clipboard sniff (Accessibility is still required for injection, unchanged).
- No smoke-test contract classes changed. Smoke tests run with `isMac=false`, so the Windows hotkey display/behavior is unaffected.

## Model Used

- Anthropic **Claude Opus 4.8** (`claude-opus-4-8`), via Claude Code with extended thinking + tool use.

## Checklist

- [x] Checked `docs/ROADMAP.md` - this PR does not duplicate planned work (it advances the macOS-2 permission issues)
- [x] Smoke tests pass (unified fast profile, 16/16)
- [x] Tests added or updated where applicable (permissions summary test)
- [ ] UI changes include before/after screenshots (macOS-only UI; not captured in this environment)
- [ ] If version bumped: all three files updated together (N/A - no version bump)
- [x] Risks documented above

### DIFF
--- a/docs/macos-code-signing.md
+++ b/docs/macos-code-signing.md
@@ -9,15 +9,18 @@ npm run tauri:build:signed
 ```
 
 This signs the app with a **stable self-signed identity** ("Open Flow Self-Signed")
-so that Accessibility and Input Monitoring permissions you grant **survive
-rebuilds**. Building with plain `npm run tauri build` (ad-hoc signing) will make
-the app forget its permissions every time you rebuild.
+so that the Accessibility permission you grant **survives rebuilds**. Building
+with plain `npm run tauri build` (ad-hoc signing) will make the app forget its
+permission every time you rebuild.
+
+> The global hotkey now uses Carbon `RegisterEventHotKey` (no Input Monitoring
+> permission), so **Accessibility** (for Cmd+V injection / AX reads) is the only
+> rebuild-sensitive TCC grant left. Microphone self-heals (see below).
 
 ## Why this matters
 
-macOS TCC (Transparency, Consent & Control) ties **Accessibility** and **Input
-Monitoring** grants to the application's *code-signature identity*, not its path
-or bundle id.
+macOS TCC (Transparency, Consent & Control) ties the **Accessibility** grant to
+the application's *code-signature identity*, not its path or bundle id.
 
 - An **ad-hoc** signature (`signingIdentity: "-"` in `tauri.conf.json`) has no
   stable identity. TCC falls back to keying the grant on the binary's `CDHash`,
@@ -27,16 +30,15 @@ or bundle id.
 
 The symptom is confusing: **System Settings → Privacy & Security → Accessibility
 shows the app as enabled, but the app still behaves as if permission is missing**
-(the global hotkey and clipboard injection silently stop working). That is
-because the System Settings list matches the entry by path for *display*, while
-the runtime `AXIsProcessTrusted()` / `IOHIDCheckAccess()` checks compare the
-*running* binary's signature against what TCC actually authorised — and they
-don't match after a rebuild.
+(clipboard injection silently stops working). That is because the System Settings
+list matches the entry by path for *display*, while the runtime
+`AXIsProcessTrusted()` check compares the *running* binary's signature against
+what TCC actually authorised — and they don't match after a rebuild.
 
-> Microphone is the exception that hid this for a while: the mic path re-prompts
-> through AVFoundation (re-establishing the grant against the current binary) and
-> also has an empirical "I actually captured audio" latch, so it self-heals.
-> Accessibility and Input Monitoring cannot re-prompt the same way.
+> Microphone is the exception: the mic path re-prompts through AVFoundation
+> (re-establishing the grant against the current binary) and also has an empirical
+> "I actually captured audio" latch, so it self-heals. Accessibility cannot
+> re-prompt the same way.
 
 Signing every local/release build with **one reused certificate** gives the
 bundle a stable designated requirement, so a single grant persists across all
@@ -102,11 +104,10 @@ is tracked separately and is out of scope here.
 
 ## Recovering from stale grants
 
-If you previously installed ad-hoc builds and the OS is holding stale entries,
+If you previously installed ad-hoc builds and the OS is holding a stale entry,
 the in-app **Settings → Permissions → "Reset stale grants"** button runs
-`tccutil reset Accessibility com.verenu.app` and `tccutil reset ListenEvent
-com.verenu.app`, then walks you through re-adding the app. After switching to
-signed builds you should only have to grant once.
+`tccutil reset Accessibility com.verenu.app`, then walks you through re-adding the
+app. After switching to signed builds you should only have to grant once.
 
 ## Runtime resilience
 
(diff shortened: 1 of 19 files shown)


MONKE2525E:
@gemini-code-assist Thanks for the review — all three issues were valid and I've addressed them in commit `a3572c0`.

**1. Modifier-less single-key hotkeys can hijack typing (`build_hotkey`)** — Fixed. `build_hotkey` now rejects a modifier-less single key unless it's a function key (F1–F12), since a bare printable key (letter/`Space`) registered via `RegisterEventHotKey` would be consumed system-wide and break normal typing. The default ⌥+Space is unaffected because it carries a modifier.

**2. Frontend hotkey capture allows bare keys (`GeneralSection`)** — Fixed. On macOS a single key is now only accepted when it matches `/^F([1-9]|1[0-2])$/` (F1–F12); any other bare key shows the invalid-key hint instead of registering immediately. This mirrors the backend rule.

**3. Escape-registration race (`ESCAPE_REGISTERED`)** — Fixed. Replaced the `AtomicBool` with a `Mutex<bool>` and moved the `register`/`unregister` calls inside the lock so the check-and-act is a single critical section. `set_escape_listening` is invoked from both the hotkey event thread and Tauri command threads, so the previous lock-free `swap` could interleave and leave Escape permanently registered.

I also added unit tests covering `build_hotkey`'s accept/reject rules (bare letter/`Space` rejected, bare F-key allowed, modifier+key allowed, modifier-only rejected).

Verification: `npm run lint` (svelte-check + `clippy -D warnings`) is clean and `cargo test` passes (216 tests).

Could you take another look?

/gemini review

MONKE2525E:
@gemini-code-assist I expanded the PR comment search and found a second batch of your review comments. I validated each one locally, and the follow-up issues were valid.

I addressed them in `62996dd`:

1. Escape is now registered only during active hold-to-talk capture, so handsfree mode no longer hijacks Escape system-wide in the background.
2. The macOS hotkey startup path now returns early when the global hotkey manager is already initialized, which prevents duplicate event-listener threads from competing for the shared hotkey event channel.
3. The macOS paste-injection error log now respects the empirical accessibility-verification latch, so a stale TCC cache will not emit a misleading Accessibility error after access has already been proven.

I also re-validated the earlier three hotkey fixes already on this branch.

Verification:
- `cargo test --manifest-path src-tauri/Cargo.toml core::hotkey::mac::tests`
- `npm run lint`

Could you take another pass and let me know if there is anything else to address?

/gemini review

MONKE2525E:
@gemini-code-assist I found and validated the next review batch.

Two code issues were valid, and I fixed them in `e302923`:

1. Startup hotkey migration now applies the migrated default in memory as well as in settings, and save failures during that migration now log a warning instead of being silently ignored.
2. macOS hotkey mapping now includes common keys like Escape, Backspace, Delete, Home, End, Page Up, and Page Down, so modifier-plus-key rebinds for those keys register correctly. I also added focused test coverage for representative combinations.

I did not restore global Escape cancellation for handsfree mode. Doing that would reintroduce the system-wide Escape hijack from the previous high-priority issue. On macOS, the safer tradeoff is:
- active hold-to-talk: Escape remains available transiently to cancel the live hold session
- handsfree mode: stop or cancel via the main hotkey or the pill UI, rather than swallowing Escape globally in every app

This comment is intended to clarify that tradeoff so the current behavior is explicit.

Verification:
- `cargo test --manifest-path src-tauri/Cargo.toml core::hotkey::mac::tests`
- `npm run lint`

Could you take another pass and let me know if anything else remains?

/gemini review

MONKE2525E:
@gemini-code-assist I found and fixed the startup Accessibility gate you just flagged.

`6a3b43e` updates the macOS recording-start check to honor the empirical `is_accessibility_verified()` latch before falling back to the raw OS trust check. That keeps stale cached `AXIsProcessTrusted()` results from incorrectly blocking recording after Accessibility has already been proven to work.

Verification:
- `npm run lint`
- `cargo test --manifest-path src-tauri/Cargo.toml core::hotkey::mac::tests`

Could you take one more pass and let me know if there is anything else left to address?

/gemini review

MONKE2525E:
@gemini-code-assist I validated the latest setup-step note and cleaned it up in `b1b4271`.

The `Try It` step now initializes from `defaultHotkey` directly, since that helper already performs the platform check internally and the extra ternary was redundant.

Verification:
- `npm run lint`

Could you take one final pass and let me know if anything else remains?

/gemini review

PR #142, opened 2026-06-22
refactor: split oversized modules into submodules + security/privacy/a11y fixes

MONKE2525E:
## Thinking Path

> - Verenu is a Windows/macOS AI dictation app — hotkey hold → audio capture → transcription → LLM cleanup → text injection.
> - This work spans several subsystems: the Rust pipeline/injection/prompts/auto_learn modules, the updater, logging, and the Svelte settings/views frontend.
> - The large modules (`pipeline.rs`, `core/injection.rs`, `api/prompts.rs`, plus `Home`/`Dictionary`/`Snippets`/`ModelsSection` Svelte views) had grown into multi-hundred/thousand-line files that were hard to navigate and review; separately, a deep review surfaced a few real bugs (log leaks, Google key in URL, privacy beacon, missing a11y labels, stale "Open Flow" rename drift).
> - It needs addressing now because the files were past the point where further feature work was painful, and the fixes are low-risk hardening that shouldn't wait behind the refactor.
> - This pull request splits the oversized modules into focused submodules and lands the accompanying bug/edge-case fixes, rebased cleanly onto the latest `dev`.
> - The benefit is a smaller, more reviewable surface per file plus several correctness/privacy/security improvements, with no intended behavior change from the refactor itself.

## What Changed

- **Module splits (Rust):** `pipeline.rs` → `pipeline/` (cache, chains, finalize, session, stages, state, fixture, tests); `core/injection.rs` → `core/injection/` (mod, windows, macos); `api/prompts.rs` → `api/prompts/` (mod, transcription, cleanup_rules, cleanup_templates, gemini, tests); `api/auto_learn` gains `focused_text`, `monitor`, `rejection` submodules plus a `correction` algorithm extraction.
- **Module splits (frontend):** `Home`, `Dictionary`, `Snippets` views split into `views/{home,dictionary,snippets}/` pieces; `ModelsSection` split into `ModelProviderGroup`/`ModelTaskTile`/`models.ts`; app mappings extracted to `components/appMappings/`; new reusable `modalFocus` helper.
- **fix(security):** redact dictation previews from logs unless verbose dev logging is on; move the Google API key from the `?key=` query string into the `x-goog-api-key` header for transcription + cleanup, and strip it in the log redactor. (The updater-RCE hardening from the original commit is already present on `dev` via #141, so only the log/key-exposure parts land here.)
- **fix(privacy):** drop the `google.com` connectivity beacon, add a CSP, and disclose the connectivity check.
- **fix(a11y):** label icon-only buttons and secret inputs for screen readers.
- **fix:** purge stale "Open Flow" / `com.openflow` rename drift left over from the Verenu rename.

## Verification

- `npm run check` (svelte-check) — **passes**, 158 files, 0 errors / 0 warnings.
- `cargo check --manifest-path src-tauri/Cargo.toml` — **passes** (Windows target).
- Branch rebased onto the latest `origin/dev` (merge-base == `origin/dev` tip); the already-merged "native update notifications" commit was dropped so the PR contains only the intended 8 commits.
- Manual conflict resolutions: dev's stricter `is_authorized_release_asset_url` and full Windows auto-update were kept; dev's `#140` macOS `is_accessibility_verified()` short-circuit and `tap_active` log removal were ported into the new `pipeline/`/`injection/` submodules.

## Risks

- **macOS code paths not compiled here.** The ported `#[cfg(target_os = "macos")]` lines (`is_accessibility_verified()` short-circuit, `tap_active` log removal) mirror dev's exact pattern and reference `mac_app::is_accessibility_verified` (which exists on `dev`), but were not type-checked by the Windows `cargo check`. A macOS build/CI run should confirm.
- **Large refactor surface** (~106 files). The module splits are intended to be behavior-preserving; the risk is import-path / wiring regressions rather than logic changes.
- `npm run lint` (ESLint + Clippy), `npm run test:rust`, and the Playwright smoke suite were **not** run locally — CI should exercise them.

## Model Used

Anthropic Claude Opus 4.8 (`claude-opus-4-8`), via Claude Code with extended thinking and tool use.

## Checklist

- [ ] Checked `docs/ROADMAP.md` - this PR does not duplicate planned work
- [ ] `npm run lint` passes (ESLint + Clippy) — not run locally
- [ ] `npm run test:rust` passes — `cargo check` passed; full test suite not run locally
- [ ] Smoke tests pass (see `Agent-Skills/SmokeTest.md` for commands) — not run locally
- [x] Tests added or updated where applicable (pipeline/prompts/auto_learn test modules carried into the split)
- [ ] UI changes include before/after screenshots — N/A (refactor is behavior-preserving)
- [x] If version bumped: all three files updated together — N/A (no version bump)
- [x] Risks documented above


MONKE2525E:
@gemini-code-assist Thanks for the review! I've pushed a fix commit addressing both findings, plus the macOS CI failure. Summary of what changed and why:

**1. (High) `AboutSection.svelte` — `installActionLabel` null safety — fixed.**
You're right that the refactor dropped the original `UpdateInfo | null` contract. In practice the only call site is guarded by `{#if updateCheckState === 'available' && appStore.updateInfo}`, and `svelte-check` narrows the store access there, so there was no live crash path. But the helper lost its defensive contract and could throw if called earlier, so I restored it exactly as you suggested:

```ts
function installActionLabel(update: UpdateInfo | null): string {
  if (!update) return 'Install Now';
  return update.installMode === 'download' ? downloadActionLabel(update) : 'Install Now';
}
```

**2. (Medium) `pipeline/stages.rs` — redundant `verenu:error` emit — fixed.**
Confirmed: `show_error_pill` already calls `log::error!` and `app.emit("verenu:error", …)` internally, so the extra `app.emit` produced a duplicate toast. I removed it and folded the more descriptive message ("…Please split it into shorter dictations.") into the single `show_error_pill` call, so the user still gets the better wording with no double notification.

**3. CI: macOS `clippy -D warnings` failure (`unused import: focused_text::*`) — fixed.**
This was a genuine platform-specific lint. The `focused_text` glob feeds `pub(super)` helpers (`classify_caret_char`, `resolve_injection_context`) to the test module via `use super::*`, and the public items are also re-exported explicitly — so the glob can't be removed without breaking the tests, but the macOS lib build uses none of its items directly. I added a scoped `#[allow(unused_imports)]` with an explanatory comment, matching the existing pattern already used on the adjacent re-export. `cargo clippy -- -D warnings` and `svelte-check` both pass locally.

Could you take another look when you have a chance? `/gemini review`

MONKE2525E:
@gemini-code-assist Thanks — good catch on the clipboard path. Fixed in the latest commit.

**(High) `write_clipboard_unicode` — `GlobalLock` null deref + leak — fixed.**
You're right: `GlobalLock(hg)` was dereferenced by `copy_nonoverlapping` with no null check, and on failure the allocated `hg` leaked while the clipboard stayed open. I added the guard:

```rust
let ptr = GlobalLock(hg) as *mut u16;
if ptr.is_null() {
    let _ = GlobalFree(Some(hg));
    CloseClipboard().ok();
    return Err(anyhow::anyhow!("GlobalLock failed"));
}
```

One small adjustment from your suggestion: this crate pins `windows = "0.61"`, where `GlobalFree` lives in `Win32::Foundation` (not `Win32::System::Memory`) and takes `Option<HGLOBAL>`, so I imported it from `Foundation` as `GlobalFree(Some(hg))`. `cargo clippy -- -D warnings` passes with the change.

Could you take another look when you get a chance? `/gemini review`

MONKE2525E:
@gemini-code-assist Thanks — all the encoding findings are fixed.

**(Medium) UTF-8 mojibake from the module split — fixed (whole class).**
You're right that the split round-tripped several files through a non-UTF-8 encoding and corrupted em-dashes (`—`), arrows (`→`), and ellipses (`…`) in both comments and user-facing strings. Rather than patch only the sampled lines, I swept every source file and repaired all of them, so this also covers a few occurrences beyond the ones you flagged (e.g. in `data/db` test comments and additional pipeline strings).

Notably the user-visible messages are restored, e.g.:
- "Audio too quiet — check your mic"
- "Recording was too quiet — nothing was transcribed"
- the macOS Accessibility hint: "…System Settings → Privacy & Security → Accessibility"

I verified zero mojibake markers remain across the tree, and `cargo clippy -- -D warnings` + `svelte-check` both pass.

Could you take another pass when convenient? `/gemini review`

MONKE2525E:
@gemini-code-assist Thanks — all four addressed in the latest commit.

**1. (High) `write_clipboard_unicode` — leak when `SetClipboardData` fails — fixed.**
Applied your suggestion: on failure I now `GlobalFree(Some(hg))` and `CloseClipboard()` before returning the error. (On success the system takes ownership of `hg`, so it's correctly only freed on the error path.)

**2. (Security-High) connectivity check exhausting the GitHub API rate limit — fixed.**
Confirmed and important: `App.svelte` polls this every 60s, which is exactly the 60/hr unauthenticated API budget and would 403-starve the updater. Switched the probe from `https://api.github.com` to `https://github.com` (still a host the app legitimately contacts for release downloads, but not the rate-limited API endpoint).

**3. (Medium) untrimmed app mappings never resolving — fixed (resolution side).**
You're right about the mismatch. I fixed it in `resolve_app_mapping` by trimming the stored `exe` before the case-insensitive compare, rather than tightening the validator. Two reasons: it also repairs any already-persisted untrimmed entries (a save-time reject only helps future writes), and it can't regress saves. Note the dedup key in the validator is already trimmed+lowercased, so trim-variant duplicates are still rejected.

**4. (Medium) close button SVG missing `aria-hidden` — fixed.**
Added `aria-hidden="true"` to match the minimize button.

`cargo clippy -- -D warnings` and `svelte-check` both pass. Could you take another look? `/gemini review`

MONKE2525E:
@gemini-code-assist Thanks — here's how I handled this round.

**1. (Critical) full dictated text/prompt/app-context logged at debug — no change needed (already guarded).**
These three `log::debug!` lines are already inside `if crate::system::logger::is_verbose() { … }` (the block opens at `cleanup.rs:57`, just above the hunk you saw). So the full text, prompt, and app context are only ever written when developer verbose logging is explicitly enabled — they don't land in default logs. I believe the diff hunk started at line 58 and the enclosing guard wasn't visible. No code change here; happy to be corrected if you're seeing a path that bypasses line 57.

**2. (Medium) untrimmed app-mapping profile/intensity silently degrading tone — fixed.**
Good catch on the `profile`/`cleanup_intensity` variant. I fixed it in `apply_app_style_overrides` by trimming both before they're used, which complements the `resolve_app_mapping` exe-trim from the previous round. I deliberately fixed it at runtime rather than rejecting on save: that way an imported backup with padded values is corrected and still works, instead of the whole import being refused, and it also repairs any already-persisted entries. The validator's dedup key is already trimmed+lowercased, so duplicate trim-variants are still rejected.

**3. (Medium) logger `key=` redaction allocation + false positives — fixed.**
Applied your approach: I now redact the `?key=` and `&key=` query markers directly and dropped the `out.to_ascii_lowercase().contains("googleapis.com")` gate. That removes the per-message `String` allocation and avoids clobbering unrelated params like `cache_key=` (which contains neither `?key=` nor `&key=`).

`cargo clippy -- -D warnings` passes. Could you take another look? `/gemini review`

PR #143, opened 2026-06-22
refactor: backend-owned settings store, dependency slimming, logging hardening

MONKE2525E:
## Thinking Path

> - Verenu is a Windows/macOS AI dictation app — hotkey hold → audio capture → transcription → LLM cleanup → text injection, with users supplying their own API keys and a ~200MB idle RAM target.
> - This touches the **settings**, **build/dependency**, **API cleanup**, and **logging** subsystems.
> - PR #142 split oversized modules and landed security/privacy/a11y fixes, but left the settings layer on `tauri-plugin-store` and carried build/runtime dependencies (PostCSS toolchain, full `tokio`, default `reqwest`, unused crates) heavier than the project's lean-footprint goals require.
> - It needs addressing now while the module boundaries from #142 are fresh, so the settings store and dependency surface settle before further feature work builds on them.
> - This pull request replaces `tauri-plugin-store` with a backend-owned JSON settings store, slims the dependency/build surface, unifies the cleanup provider abstraction, and hardens log redaction.
> - The benefit is a smaller dependency footprint (supporting the RAM target), a settings layer the backend fully owns and validates, and stronger guarantees that dictated text never reaches logs.

## What Changed

- **Backend-owned settings store**: replaced `tauri-plugin-store` with a JSON-backed `SettingsHandle`/`SettingsSnapshot` in `data/store.rs`; added a declarative `SettingSpec`/`SettingKind` validation table in `commands/settings.rs` for readable/exportable per-key handling.
- **Dependency & build slimming**: dropped `@tailwindcss/postcss`, `autoprefixer`, `esbuild`, `postcss` devDeps and removed `tailwind.config.js` / `postcss.config.js` (Tailwind v4 CSS-based config); trimmed `tokio` from `full` to `rt`/`rt-multi-thread`/`macros`/`sync`/`time`; set `reqwest` to `default-features = false` + `native-tls`/`charset`; removed `tauri-plugin-store` and unused macOS crates (`foreign-types-shared`, `core-foundation-sys`); simplified the vite `minify` setting.
- **Unified cleanup provider**: folded the `CleanupProvider` enum into the shared `ProviderId`, with provider-driven endpoint/config (`cleanup_url()`) and shared Gemini request types.
- **Logging/privacy hardening**: expanded the redaction token list (`raw_full=`, `input_full=`, `prompt_full=`, `clipboard=`, etc.) and added `sanitize_frontend_log_message()` to truncate frontend-supplied log messages to 1,000 chars.

## Verification

- `npm run check` (TypeScript) and `npm run lint` (ESLint + Clippy).
- `npm run test:rust` — covers settings store and prompt/cleanup paths.
- Manual settings round-trip: change provider/tone/mic-gain/app-mappings in Settings → confirm persistence across app restart (settings.json) and that API-key presence checks still return booleans only.
- Pipeline smoke: hotkey hold → dictation → injection, confirming cleanup provider routing still works for Groq/OpenAI/Google.
- Build check: `npm run tauri build` succeeds without the removed PostCSS toolchain.

## Risks

- **Settings migration**: existing users move from `tauri-plugin-store`'s store file to the backend-owned `settings.json`. Verify no settings are lost on upgrade for users with an existing install.
- **Dependency feature trimming**: narrowed `tokio`/`reqwest` features could surface a missing-feature compile/runtime error on a path not exercised by tests; CI OS-matrix Rust build is the gate.
- API keys remain in the OS credential store — unchanged; settings.json still never holds secrets.
- Low risk to clipboard/injection and hotkey timing — those paths are untouched.

## Model Used

- Anthropic Claude (Opus 4.8) via Claude Code, with tool use and extended context.

## Checklist

- [ ] `npm run check` passes (TypeScript)
- [x] If version bumped: all three files updated together (no version bump in this PR)
- [ ] Risks documented above


MONKE2525E:
@gemini-code-assist Thanks for the review — both findings were valid and I've addressed them in `3d3326f`.

**1. (High) Corrupted `settings.json` crashing startup**
`read_settings_file` no longer propagates a parse error. On a parse failure it now logs the error, backs the bad file up to `settings.json.bak`, and falls back to an empty settings map so the app still launches. I extended this slightly beyond the suggestion: the *valid-JSON-but-not-an-object* case took the same `Err` → startup-crash path, so it now backs up and resets too. Both cases route through a small `backup_corrupt_settings` helper.

**2. (Medium) Non-atomic write truncating settings**
`write_settings_file` now serializes to `settings.json.tmp` and then atomically renames it over the live file, so an interrupted write (crash, power loss, disk full) can't leave a truncated `settings.json`. On Windows `std::fs::rename` maps to a replace-existing move, so the swap is atomic there as well.

The crate builds clean (`cargo build`). Could you take another look when you have a chance?

MONKE2525E:
@gemini-code-assist Thanks for the follow-up — both points were valid and are fixed in `2fff10e`.

**1. (Medium) Silent lock poisoning in `SettingsHandle::get`**
Replaced the `.ok()` swallow with an explicit `match`. On a poisoned lock it now logs an error (including the key being read) and returns `None`, so the inconsistent in-memory state is diagnosable rather than silently masquerading as a missing setting.

**2. (Medium) Temp file leaking on rename failure in `write_settings_file`**
The rename is now handled explicitly: if it fails, the `settings.json.tmp` file is removed before returning the error, so repeated failures (file locks, permissions) can't accumulate temp files in AppData.

Crate builds clean (`cargo build`). Could you take another look?

MONKE2525E:
@gemini-code-assist Thanks — all four were valid and are fixed in `cb133e3`. I confirmed the constants exist in `store.rs` (`HOTKEY` and `SETUP_COMPLETE`) and that the raw literals were present at the flagged lines before replacing them:

- `main.rs:156` — `settings.get("hotkey")` → `settings.get(crate::data::store::HOTKEY)`
- `main.rs:170` — `.set("hotkey", ...)` (macOS hotkey migration) → `.set(crate::data::store::HOTKEY, ...)`
- `main.rs:210` — `.get("setup_complete")` → `.get(crate::data::store::SETUP_COMPLETE)`
- `commands/system.rs:66` — `save_value("hotkey", ...)` → `save_value(store::HOTKEY, ...)`

These were the last leftover raw key literals in those two files (other call sites already used the constants). Crate builds clean. Could you take another look?

MONKE2525E:
@gemini-code-assist Both addressed in `f37b089`.

**1. Corrupt-settings backup reliability**
Applied your suggestion: `backup_corrupt_settings` now removes any existing backup before the rename and logs at `warn` (it's a non-critical startup cleanup). One honest note: Rust's `std::fs::rename` on Windows maps to `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING`, so it already replaces an existing destination — the unconditional `remove_file` is belt-and-suspenders for filesystems where replace-on-rename isn't guaranteed, plus it makes the intent explicit. No harm, and the `error`→`warn` downgrade is the right call regardless.

**2. Bounded redaction in `sanitize_frontend_log_message`**
Valid CPU concern. The input is now capped (to `MAX_FRONTEND_LOG_CHARS * 2`, only when the message exceeds `* 4`) before the redaction passes run, so an oversized frontend message can't drive repeated case-insensitive scans over megabytes. The headroom keeps a sensitive token from being sliced at the boundary, and the final output cap is unchanged.

Crate builds clean. Could you take another look?

MONKE2525E:
@gemini-code-assist Addressed in `417cbed`.

**Blocking I/O on the async executor (4 commands)** — valid, fixed. `save_app_mappings`, `save_setting`, `save_hotkey`, and `set_autostart` now wrap the synchronous `save_value` write in `run_blocking(...)`, matching the pattern already used by the other commands in `library.rs`. They no longer block the Tokio threadpool.

**UTF-8 cursor advance in `redact_quoted_field_value_ci`** — I applied your char-boundary advancement, but with one correction for the record: this was **not** an actual panic risk in the current code. The scan function `find_ascii_case_insensitive_from` doesn't slice on the cursor — it iterates byte indices and `continue`s on `!haystack.is_char_boundary(i)`, and the only `replace_range`/`find_json_string_end` slices use boundaries derived from an ASCII `"`. So a mid-char cursor would have been skipped safely rather than panicking. That said, advancing by a whole char is cleaner and removes the latent fragility if that scan function ever changes, so I took the fix.

Crate builds clean. Could you take another look?

MONKE2525E:
@gemini-code-assist Fixed in `7ae0691`. `write_settings_file` now removes the `settings.json.tmp` file if the `std::fs::write` itself fails (full disk, permissions), before returning the error — so a failed or partial write can't leave a stale temp file behind. This mirrors the rename-failure cleanup added earlier, so both failure paths in the atomic-write now clean up after themselves.

Crate builds clean. Could you take another look?

MONKE2525E:
@gemini-code-assist I looked into this one carefully and I'm respectfully declining the suggested change, because I believe the premise is incorrect and the proposed fix would actually regress the atomicity guarantee.

**Why the premise doesn't hold:** Rust's `std::fs::rename` is documented to *replace* an existing destination on every platform. On Windows it is implemented with `MoveFileExW` / `SetFileInformationByHandle` using replace-existing semantics — it is **not** the C `rename()` / raw `MoveFile` behavior that fails when the target exists. So overwriting an existing `settings.json` via temp-file + rename works on Windows; saves after the first do not fail.

**Evidence:** I added a regression test (`write_settings_file_overwrites_existing`) in `d4f8bee` that saves to a path, then saves again to the same path, and asserts the second save succeeds and the new contents are read back. It passes on Windows (and runs in the OS-matrix CI). I also added `read_settings_file_recovers_from_corrupt_json` to lock in the corrupt-file backup behavior.

**Why I didn't add the `#[cfg(windows)] remove_file(path)` before the rename:** that would delete the live `settings.json` first and then rename, opening a window where a crash/power-loss between the two leaves the user with *no* settings file at all. That's the exact non-atomic failure mode the temp-file + rename pattern was added to prevent, so it would be a step backward.

The temp-file cleanup on the genuine failure paths (write error, rename error) is already in place from the previous rounds. Happy to revisit if you can point to a concrete case where `std::fs::rename` does not replace on a supported Windows target. Could you take another look?

Issue #144, opened 2026-06-22
App mapping override can leak globally across apps


MONKE2525E:
## Problem
A per-app mapping was configured for one app with **Very Casual** tone and **Direct** cleanup, but the behavior appeared to become global. The mapping had to be deleted because dictation in other apps started coming out with the same very-casual/direct style.

This may not be specific to the Very Casual + Direct combination, but that combination made the leak obvious because the output style was very noticeable.

## User-visible behavior
- Create an app mapping for one app.
- Set tone to **Very Casual**.
- Set cleanup intensity to **Direct**.
- Dictate into other apps that should use the global/default tone and cleanup level.
- Output still appears to follow the mapped app's Very Casual + Direct behavior.

## Expected behavior
- A mapping should only affect dictation when the active foreground app exactly matches that mapping.
- When no mapping matches, Verenu should fall back to the global `default_tone` and global `cleanup_intensity`.
- Deleting one app mapping should never be required to restore global behavior in unrelated apps.

## Research notes / likely code areas
- `src/lib/appMappings.ts`
  - Defines mapping shape: `exe`, `profile`, optional `cleanup_intensity`.
  - Normalizes executable names with `normalizeExe()`.
- `src/lib/components/AppMappingsEditor.svelte`
  - Creates and saves mappings.
  - Defaults new mappings to `profile: 'casual'` and `cleanup_intensity: ''` unless overridden.
- `src-tauri/src/pipeline.rs`
  - `resolve_app_mapping()` matches the active `process_name` against saved mappings.
  - `apply_app_style_overrides()` mutates `cfg.cleanup_intensity` if a mapping has an override, and returns the mapped profile or `cfg.default_tone`.

## Things to verify
- Log the active `process_name`, matched mapping exe/name, final `profile`, and final `cleanup_intensity` for every dictation.
- Confirm `resolve_app_mapping()` returns `None` when dictating into unrelated apps.
- Confirm `process_name` is not accidentally resolving to a generic value like `unknown`, `verenu`, browser host process, or stale previous process.
- Confirm app mappings are not accidentally saved into global tone/cleanup settings by the UI.
- Confirm retry/handsfree paths do not reuse stale `profile`, `app_context`, or `cleanup_intensity` from a previous app.

## Acceptance criteria
- Add coverage or instrumentation proving app mapping overrides are scoped to the matching executable only.
- Add a regression test for: mapping exists for App A, active process is App B, effective tone/intensity remain global defaults.
[...]

PR #153, opened 2026-06-23
fix: app-mapping style leak (#144) + contextual caps for empty/Chromium text boxes

MONKE2525E:
## Thinking Path

> - This touches two subsystems: the **pipeline** (per-app tone/cleanup mapping) and **text injection** (contextual capitalization + spacing)
> - Two everyday-correctness gaps existed: a per-app mapping could leak its style into other apps, and dictation into empty or Chromium/Electron text boxes (e.g. Antigravity's Monaco composer) came out lowercased with no leading space because UI Automation reports phantom text before the caret
> - These needed addressing now because they made app tone-mappings effectively unusable (issue #144) and made everyday dictation output look wrong in real editors — directly in scope of ROADMAP §3 "Contextual Capitalization Reliability Regression"
> - This pull request anchors mapping resolution to the window text is actually injected into, and adds a Windows clipboard sniff that verifies the real box contents before deciding capitalization, while keeping spacing on the reliable UIA signal
> - The benefit is that tone mappings stay scoped to their app, and dictation capitalizes and spaces correctly even where the accessibility layer lies — and it fails closed (never randomly lowercases), advancing the deterministic-context goal in ROADMAP §3

## What Changed

- **App-mapping style leak (#144):** the per-app tone + cleanup-intensity now resolve from the captured injection-target window (`target_hwnd`) instead of the live foreground at hotkey-release. The two could diverge (handsfree, focus shifts), which let one app's Very Casual/Direct style leak into dictation injected into other apps. Falls back to the live foreground only when the captured hwnd is unavailable.
- **Casual/formal cleanup now guarantees terminal punctuation**, so a following dictation into the same field reads as a new sentence and capitalizes (`very_casual`/verbatim intensity are deliberately exempt).
- **Windows clipboard sniff for contextual caps:** when a caret-local UIA read would lowercase, it selects one character back and inspects what is *actually* selectable. An empty box now capitalizes; genuine mid-sentence joins still lowercase. The sniff drives **only** capitalization — **spacing follows the reliable UIA tail**, so appends keep their leading space even in Chromium where synthetic `Ctrl+C` is a no-op. The caret is **always** restored, so a failed read can never corrupt text.
- **Redacted diagnostics** (verbose-only): mapping resolution (process, matched exe, profile, intensity) and the capitalization decision (probe source, previous-character class, case decision) — the metadata called for in ROADMAP §3.
- **Tests:** mapping-scope regression tests, terminal-punctuation tests, and a unit-tested gate for when sniff verification runs.

## Verification

- `npm run test:rust` (`cargo test --manifest-path src-tauri/Cargo.toml`) — **246 passed**, includes new mapping, punctuation, and sniff-gate tests
- `cargo clippy --manifest-path src-tauri/Cargo.toml -- -D warnings` — clean
- `npm run check` (svelte-check) — 0 errors / 0 warnings (no frontend changes)
- Manual (Windows):
  - Map App A → Very Casual/Direct, dictate into App B → output stays the global tone (verbose log shows `matched=none`); deleting the mapping is no longer required.
  - Type `Hello`, dictate `Hi` immediately after the `o` → `Hello Hi` (capitalized, space added, **no character deleted**).
  - Dictate into a visually empty Antigravity composer → first letter capitalized.

## Risks

- **Injection behavior shift (Windows only):** the sniff sends synthetic `Shift+Left` / `Ctrl+C` / `Right` and touches the clipboard, but only on lowercase-candidate dictations. It guards on the foreground HWND, restores the caret **unconditionally**, and restores the clipboard via the existing save/restore path, so it cannot corrupt text. Worst case is an append occasionally missing its leading space, or a truly-empty Chromium box getting one leading space — both cosmetic. It adds ~150-220 ms to mid-sentence dictations. Turning off contextual capitalization disables the sniff entirely.
- **macOS path unchanged.** No settings migration, no API-key surface, no smoke-test contract changes. All new logs are redacted metadata only (no dictated text).

## Model Used

- Anthropic Claude Opus 4.8 (`claude-opus-4-8`), extended thinking + tool use

## Checklist

- [x] Checked `docs/ROADMAP.md` - this PR advances ROADMAP §3 (contextual capitalization) rather than duplicating planned work
- [x] `npm run lint` passes (ESLint + Clippy) — Clippy `-D warnings` clean; no frontend/TS lint changes
- [ ] Smoke tests pass (not run for this backend-only injection change)
- [ ] UI changes include before/after screenshots (no UI changes)
- [ ] If version bumped: all three files updated together (no version bump)
- [ ] Relevant documentation updated to reflect changes (behavior already described generically in CLAUDE.md injection section)
- [x] Risks documented above

### DIFF
--- a/src-tauri/src/api/auto_learn/focused_text.rs
+++ b/src-tauri/src/api/auto_learn/focused_text.rs
@@ -383,6 +383,19 @@ impl FocusedTextReader {
             }
             let selection_state = describe_selection_state(range_seen, range_collapsed);
 
+            // Structural diagnostics for "blank box reads as full" (Antigravity /
+            // Chromium controls). App-control metadata only — never tail content.
+            log::debug!(
+                "injection: probe detail control_type={control_type} class={class_name} automation_id={automation_id} value_empty={value_is_empty:?} read_only={read_only:?} caret_active={range_seen} collapsed={range_collapsed} source={} context={} tail_len={} tail_newline={}",
+                source.as_str(),
+                context.as_str(),
+                caret_context.as_deref().map(|t| t.chars().count()).unwrap_or(0),
+                caret_context
+                    .as_deref()
+                    .map(|t| t.contains('\n') || t.contains('\r'))
+                    .unwrap_or(false),
+            );
+
             InjectionContextProbe {
                 context,
                 source,
--- a/src-tauri/src/core/injection/mod.rs
+++ b/src-tauri/src/core/injection/mod.rs
@@ -104,6 +104,40 @@ mod tests {
     }
 
     #[test]
+    fn sniff_verification_gate_only_for_caret_local_mid_sentence() {
+        use crate::core::text_context::SentenceContext;
+        // The one case that needs verification: caps on, caret-local, mid-sentence.
+        assert!(caret_local_needs_sniff_verification(
+            true,
+            ContextProbeSource::CaretLocal,
+            SentenceContext::MidSentence,
+        ));
+        // A caret read that already says "new sentence" capitalizes — no sniff.
+        assert!(!caret_local_needs_sniff_verification(
+            true,
+            ContextProbeSource::CaretLocal,
+            SentenceContext::NewSentence,
+        ));
+        // Empty-field and history sources are already trustworthy.
+        assert!(!caret_local_needs_sniff_verification(
+            true,
+            ContextProbeSource::EmptyField,
+            SentenceContext::MidSentence,
+        ));
+        assert!(!caret_local_needs_sniff_verification(
+            true,
+            ContextProbeSource::HistoryFallback,
+            SentenceContext::MidSentence,
+        ));
+        // Contextual caps off: never sniff.
+        assert!(!caret_local_needs_sniff_verification(
+            false,
+            ContextProbeSource::CaretLocal,
+            SentenceContext::MidSentence,
+        ));
+    }
+
+    #[test]
     fn uppercase_first_word_handles_prefix_symbols() {
         assert_eq!(uppercase_first_word("hello world"), "Hello world");
         assert_eq!(uppercase_first_word("\"hello world"), "\"Hello world");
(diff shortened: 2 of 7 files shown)


MONKE2525E:
@gemini-code-assist Thanks for the review — I addressed your finding.

**`read_clipboard_text` blocking the executor (medium):** Validated and fixed. You're right that it ran a synchronous `std::thread::sleep` backoff on the `OpenClipboard` retry path while being called from the async clipboard sniff, which blocks the Tokio executor thread. I made the function `async` and switched the retry backoff to `tokio::time::sleep().await`, with the Win32 calls kept in narrow `unsafe` blocks and the call site updated to `.await`. This also matches the existing `copy_to_clipboard` helper in the same module, which already used the async + `tokio::time::sleep` pattern.

**One thing I intentionally left out of scope:** the sibling clipboard helpers (`save_clipboard_all` / `restore_clipboard_all`) use the same synchronous retry-sleep pattern, but they're pre-existing code not touched by this PR, so I kept this change tight to the function you flagged rather than widening the diff. Happy to convert them in a follow-up if you think it's worth it.

`cargo clippy -- -D warnings` and the full Rust test suite (246 tests) both pass after the change.

Could you take another look? `/gemini review`

MONKE2525E:
@gemini-code-assist Thanks for the second pass — addressed your finding.

**`ensure_terminal_punctuation` hardcoding the Western period for CJK text (medium):** Validated and fixed. The function runs on every casual/formal cleanup regardless of the transcription language, so a Chinese or Japanese dictation ending on a bare word would get a Western "." appended — out of place in that script. Added an `is_cjk` check (CJK Unified Ideographs, Extension A, Hiragana, Katakana) and append the full-width "。" instead when the last character falls in those ranges; otherwise it's unchanged. Added test coverage for Chinese, Japanese, and the "already terminated with 。" no-op case.

`cargo clippy -- -D warnings` and the full Rust test suite (247 tests) pass after the change.

Could you take another look? `/gemini review`

PR #155, opened 2026-06-23
fix(pill): even visualizer bar widths across monitor DPIs

MONKE2525E:
## Thinking Path

> - Verenu is a Windows AI dictation app — hotkey hold -> audio capture -> transcription -> LLM cleanup -> text injection.
> - Subsystem: UI — the floating dictation pill window (`src/PillApp.svelte`) and its audio-level visualizer (the row of 12 bars shown while recording / in handsfree).
> - Problem: the bars rendered at uneven widths — some 3px, some 4px on the same row, "a few in the middle" off — on a monitor whose display scale differed from the one the pill first appeared on.
> - Why now: the recently-merged target-monitor positioning work (#154) lets the pill appear on secondary monitors, which surfaced this cross-DPI rendering bug.
> - Root cause: bar width/gap are snapped to whole device pixels, but the snap was computed in a Svelte `$:` statement that read `devicePixelRatio` only through a closure. Svelte's legacy `$:` dependency tracking is syntactic, so it never marked `devicePixelRatio` a dependency and froze the value at the first monitor's scale — a fractional CSS width (e.g. `3.2px`) that then rounds unevenly on a different-scale display.
> - This pull request makes the snap genuinely reactive to `devicePixelRatio` and refreshes it live so it always tracks whichever monitor the pill is currently on.
> - The benefit is uniform-width visualizer bars on every display, at any (including fractional) scale.

## What Changed

- Pass `dpr` explicitly into `snap()` so the `barW` / `barGap` reactive statements name it as a dependency and recompute when the device pixel ratio changes (previously frozen at mount-time scale).
- Refresh `devicePixelRatio` live — each animation frame and on entering recording/handsfree — plus a self-rearming `matchMedia('(resolution: …dppx)')` watcher, because WebView2 does not reliably fire DPI-change events when the pill moves across monitors.
- Drive bar width, gap, and height from the snapped values (`--bar-w` / `--bar-gap` CSS variables and `snap(h, dpr)`) in both the recording and handsfree visualizers.

## Verification

- `npm run check` passes (TypeScript / svelte-check, 0 errors).
- `npm run test:smoke:state` passes (no smoke contract asserts the `.bar` elements; this is a regression guard).
- Modeled the device-pixel rounding end-to-end: a frozen `3.2px` CSS bar on a 100%-scale monitor rounds to `4 3 3 3 3 4 3 3 3 3 4 3` device px (uneven, matching the report); with `dpr` correctly resolved the same row becomes all `3` (even).
- Confirmed at the Svelte compiler level: the buggy form compiles to `legacy_pre_effect(() => {}, …)` (no dependency on `dpr` → runs once), the fixed form to `legacy_pre_effect(() => ($.get(dpr)), …)` (tracks `dpr` → recomputes).
- Manual: `npm run tauri dev`, then hold Ctrl+Win (recording) and trigger handsfree on a primary 1440p @ 125% and a secondary 1080p @ 100% — bars are uniform width on both, and stay even when the pill relocates between monitors.

## Risks

- Low risk. Frontend-only, scoped to `src/PillApp.svelte`; no Rust, pipeline, injection, or settings changes. The `.bar` class/structure and pill dimensions are unchanged, so the smoke-test contract is preserved. `refreshDpr()` reassigns `dpr` only when it actually changes, so the per-frame call adds no reactive churn. Worst realistic case if `window.devicePixelRatio` were ever misreported is bars equally slightly soft rather than differently sized — strictly better than the current differential rounding.

> Check [`docs/ROADMAP.md`](../docs/ROADMAP.md) before opening feature PRs — work that overlaps with planned core changes may need to be redirected. See [`CLAUDE.md`](../CLAUDE.md) for architecture and contribution guidance.

## Model Used

- Anthropic Claude Opus 4.8 (`claude-opus-4-8`) via Claude Code — extended thinking + tool use.

## Checklist

- [ ] `npm run lint` passes (ESLint + Clippy) — ran `npm run check`; full Clippy not run locally (no Rust changed). CI covers it.
- [ ] `npm run test:rust` passes — N/A, no Rust changes (CI covers it).
- [x] Smoke tests pass (`npm run test:smoke:state`)
- [ ] Tests added or updated where applicable — N/A (no automated contract for the visualizer bar geometry).
- [ ] UI changes include before/after screenshots — sub-pixel rendering fix; only visible at native resolution on a fractional-DPI monitor. Behavior described under Verification.
- [x] If version bumped: all three files updated together — N/A, no version bump
- [ ] Relevant documentation updated to reflect changes — N/A
- [x] Risks documented above

### DIFF
--- a/src/PillApp.svelte
+++ b/src/PillApp.svelte
@@ -338,7 +338,7 @@
     </div>
 
   {:else if state === 'handsfree'}
-    <div class="pill handsfree" class:dying={dying} class:hf-expanded={showHfButtons && !dying} class:no-anim={prevState === 'recording'}>
+    <div class="pill handsfree" class:dying={dying} class:hf-expanded={showHfButtons && !dying} class:no-anim={prevState === 'recording'} style="--bar-w:{barW}px; --bar-gap:{barGap}px">
       {#if showHfButtons}
         <button class="hf-btn cancel" onclick={cancelHandless} aria-label="Cancel">
           <svg width="11" height="11" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.2" stroke-linecap="round">
@@ -346,7 +346,7 @@
           </svg>
         </button>
       {/if}
-      <div class="bars-hf" style="--bar-w:{barW}px; --bar-gap:{barGap}px">
+      <div class="bars-hf">
         {#each barHeights as h, i (i)}
           <div class="bar" style="height: {snap(h, dpr)}px"></div>
         {/each}
@@ -512,9 +512,10 @@
   }
   .pill.error.err-open .err-text { opacity: 1; }
 
-  /* Handsfree: starts compact (mirrors recording), expands to 112px after 450ms */
+  /* Handsfree: starts compact (mirrors recording — same DPI-snapped width so the
+     recording→handsfree continuation doesn't jump), expands to 112px after 450ms */
   .pill.handsfree {
-    width: 72px;
+    width: calc(12 * var(--bar-w, 3px) + 11 * var(--bar-gap, 2px) + 14px);
     padding: 0 7px;
     gap: 2px;
     transition: width 0.2s cubic-bezier(0.34, 1.56, 0.64, 1),
--- a/src/PillApp.svelte
+++ b/src/PillApp.svelte
@@ -51,7 +51,9 @@
   function armDprWatch() {
     if (typeof window === 'undefined') return;
     mq?.removeEventListener('change', onDprChange);
-    mq = window.matchMedia(`(resolution: ${window.devicePixelRatio}dppx)`);
+    // Use the already-fallback-guarded `dpr` (set right before every call) so a
+    // falsy devicePixelRatio can't produce `(resolution: undefineddppx)`.
+    mq = window.matchMedia(`(resolution: ${dpr}dppx)`);
     mq.addEventListener('change', onDprChange);
   }
 
@@ -409,12 +411,14 @@
   /* Skip entry animation for seamless continuations from recording */
   .pill.no-anim { animation: none; }
 
-  /* Recording: snug wrap — 12 bars × 3px + 11 gaps × 2px + 14px padding = 72px */
+  /* Recording: snug wrap — 12 bars + 11 gaps + 14px padding. Width is derived
+     from the snapped --bar-w/--bar-gap so the wrap stays snug at fractional DPI
+     (where snapped bars sum to more than the integer-scale 72px). */
   /* 0.25s delay keeps it invisible during a fast double-click handsfree activation */
   .pill.recording {
     gap: var(--bar-gap, 2px);
     padding: 0 7px;
-    width: 72px;
+    width: calc(12 * var(--bar-w, 3px) + 11 * var(--bar-gap, 2px) + 14px);
     animation: pillIn 0.22s cubic-bezier(0.34, 1.56, 0.64, 1) 0.25s both;
   }
 
--- a/src/PillApp.svelte
+++ b/src/PillApp.svelte
@@ -16,6 +16,38 @@
 
   const BARS = 12;
 
+  // Snap CSS-px lengths to a whole number of device pixels. On fractional DPI
+  // scaling (e.g. 1.25×/1.5×) a hardcoded 3px bar maps to a fractional device
+  // width, and Chromium rounds each bar's edges independently — so neighboring
+  // bars rasterize at different physical widths and the row looks uneven. When
+  // every bar width/gap is an exact integer device-px, all edges land on the
+  // grid and every bar renders identically.
+  //
+  // `dpr` MUST track the monitor the pill is actually on. The pill window is
+  // created once and reused, so it first mounts on the user's primary display
+  // and is later moved to other monitors (which may have a different scale).
+  // A stale dpr is worse than none: snapping 3px with the *wrong* dpr produces
+  // a fractional CSS width (e.g. 3.2px) that then rounds unevenly at the real
+  // scale. WebView2 does not reliably fire matchMedia on a cross-monitor DPI
+  // change, so we also re-read devicePixelRatio live each animation frame
+  // (see refreshDpr / animateBars).
+  //
+  // `dpr` is passed explicitly into snap() so the reactive statements below name
+  // it as a dependency. Svelte's `$:` tracking is syntactic — if `snap` closed
+  // over `dpr` instead, `barW`/`barGap` would be computed once and never update
+  // when the pill moves to a monitor with a different scale.
+  let dpr = typeof window !== 'undefined' ? (window.devicePixelRatio || 1) : 1;
+  const snap = (px: number, d: number) => Math.round(px * d) / d;
+  $: barW = snap(3, dpr);
+  $: barGap = snap(2, dpr);
+
+  // Re-read the live devicePixelRatio; assigning only on change keeps Svelte
+  // from re-running barW/barGap (and the bar template) every single frame.
+  function refreshDpr() {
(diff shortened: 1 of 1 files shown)


MONKE2525E:
Thanks for the review! I validated the one issue you raised and pushed a fix in `03adbf5`.

### Stale Media Query Watcher (medium) — fixed ✅

Your analysis was correct. The `matchMedia` watcher (`mq`, `onDprChange`, `armDprWatch`) was scoped inside `onMount`, so `refreshDpr()` — which lives at component scope and runs from the animation-frame loop — could update `dpr` but had no way to re-arm the watcher. Since a `(resolution: Xdppx)` query only fires when *that specific* resolution's match state flips, once `dpr` moved via the poll the watcher stayed bound to the stale query and would miss subsequent DPI changes.

What I changed:
- Hoisted `mq`, `onDprChange`, and `armDprWatch` to component scope.
- `refreshDpr()` now calls `armDprWatch()` whenever `dpr` changes, so the watcher is always rebound to the current monitor's resolution.
- Added a `typeof window === 'undefined'` guard in `armDprWatch` for safety.
- `onMount` just performs the initial arm; the existing cleanup still removes the listener on unmount.

One bit of context on the design that I kept intentionally: the per-frame `refreshDpr()` poll remains the **primary** mechanism for tracking the active monitor's scale, because WebView2 doesn't reliably emit DPI-change events on a cross-monitor move — that unreliability is the reason the live poll exists. With this fix the `matchMedia` watcher is now a correctly-synced **backup** rather than a watcher that silently goes stale, so the two mechanisms stay consistent.

Verification: `npm run check` passes (0 errors) and the state smoke test passes. The change is scoped to the pill visualizer logic; bar class names, structure, and pill dimensions are unchanged.

@gemini-code-assist could you take another look and let me know if you have any further feedback?

MONKE2525E:
Thanks again — both points addressed in `6864a2b`.

### 1. `window.devicePixelRatio` in the media query → use `dpr` (medium) — fixed ✅
Valid. `armDprWatch()` is always called right after `dpr` is set from `window.devicePixelRatio || 1` (initial mount, `onDprChange`, and `refreshDpr`), so `dpr` is both current and fallback-guarded at every call site. Switched the query to `` `(resolution: ${dpr}dppx)` `` so a falsy `devicePixelRatio` can't produce `(resolution: undefineddppx)`.

### 2. Hardcoded `.pill.recording` width vs snapped bars (medium) — fixed ✅
Good catch on the invariant. The width was `72px`, which only equals `12·3 + 11·2 + 14` at integer DPI; once the bars are snapped, the content sums larger at fractional scales (e.g. ~64.8px of bars at 1.25×). I replaced it with:

```css
width: calc(12 * var(--bar-w, 3px) + 11 * var(--bar-gap, 2px) + 14px);
```

so the capsule always wraps the snapped bars with the intended uniform 7px padding at any scale.

One small clarification for the record: because the pill uses `box-sizing: border-box`, the bars didn't actually spill outside the rounded background before this — the overflow stayed inside the 72px border-box and instead shrank the side padding (from 7px toward ~3.6px at 1.25×). So the visible symptom was reduced/uneven padding rather than bars escaping the capsule. The `calc()` fix restores the snug-wrap design intent either way, so I applied it as suggested.

Verification: `npm run check` passes (0 errors) and the state smoke test passes. Changes remain scoped to the pill visualizer.

@gemini-code-assist please take another look when you can — let me know if there's anything else.

MONKE2525E:
Thanks — fixed in `ce11650`.

### Handsfree compact width mismatch on fractional DPI (high) — fixed ✅
Correct, and it lines up with the existing "starts compact (mirrors recording)" intent plus the `no-anim` seamless continuation when `prevState === 'recording'`. `.pill.handsfree` now uses the same `calc(12 * var(--bar-w) + 11 * var(--bar-gap) + 14px)` width as recording, so the compact size matches at every scale and the recording→handsfree transition no longer jumps before expanding to 112px.

Implementation note: the `--bar-w` / `--bar-gap` custom properties were being set on `.bars-hf` (the child). A `calc()` width on `.pill.handsfree` can't read vars defined on a descendant, so I moved them up onto the `.pill.handsfree` element — `.bars-hf` and `.bar` still pick them up via normal inheritance.

One deviation from the snippet, for transparency: I kept `.pill.handsfree { gap: 2px }` rather than switching it to `var(--bar-gap)`. That `gap` is the **flex gap between the cancel button / bars / confirm button**, not the gap between bars. While unexpanded there's only one child (the bars container) so it has no effect, and once expanded `.pill.handsfree.hf-expanded { gap: 4px }` overrides it. The actual inter-bar gap is already `var(--bar-gap)` on `.bars-hf`. So changing the pill-level gap to the bar gap would conflate two different spacings without any visual effect — happy to change it if you'd prefer the consistency, but I left it semantically distinct.

Verification: `npm run check` passes (0 errors) and the state smoke test passes.

@gemini-code-assist could you take another look? Thanks for the thorough pass on the DPI edge cases.

MONKE2525E:
Thanks — all three addressed in `3b76929`.

### 1. `window` guard in `refreshDpr` (medium) — fixed ✅
Added `if (typeof window === 'undefined') return;` at the top, matching the guards already in `armDprWatch` and the `dpr` initializer.

### 2. `@keyframes processIn` hardcoded `72px` (medium) — fixed ✅
Good call — this was the recording→processing counterpart to the earlier width fix. As you noted, the custom properties were only on the pill elements, so the processing pill couldn't inherit them. I hoisted `--bar-w`/`--bar-gap` up to the `.wrap` container (removing the per-pill inline copies), so every state — including the bar-less processing pill — inherits the snapped values. `processIn` now starts from `calc(12 * var(--bar-w) + 11 * var(--bar-gap) + 14px)`.

I also applied the same treatment to `@keyframes processFromHf` (handsfree→processing), which had the identical hardcoded `112px` start and would jump once the expanded handsfree width became dynamic (see #3). It now starts from the `+ 54px` expanded width.

### 3. Expanded handsfree `112px` overflow (medium) — fixed ✅
Correct, and this was the more serious one since `flex-shrink: 0` bars would push the buttons out of the capsule. `.pill.handsfree.hf-expanded` is now `calc(12 * var(--bar-w) + 11 * var(--bar-gap) + 54px)`, where 54px = two 18px buttons + 10px padding + two 4px gaps. At integer DPI this still resolves to 112px; at fractional DPI it grows to fit the snapped bars exactly.

Net effect: every pill width (recording, handsfree compact + expanded) and every width keyframe now derives from the same snapped `--bar-w`/`--bar-gap`, so widths stay consistent and transition-jump-free at any DPI.

Verification: `npm run check` passes (0 errors) and the state smoke test passes.

@gemini-code-assist please take another look — appreciate the careful sweep through the fractional-DPI cases.

MONKE2525E:
Gemini Code Assist has signed off with no further feedback. Summary of the review pass for the record — all seven points were validated against the code and fixed:

1. Stale `matchMedia` watcher → hoisted to component scope, re-armed on every `dpr` change (`03adbf5`)
2. Unsafe `window.devicePixelRatio` in the media query → use the fallback-guarded `dpr` (`6864a2b`)
3. Hardcoded `.pill.recording` width → `calc()` from snapped bar vars (`6864a2b`)
4. Handsfree compact width mismatch → same `calc()` as recording (`ce11650`)
5. `refreshDpr` missing `window` guard → added (`3b76929`)
6. `processIn`/`processFromHf` keyframe jumps → vars hoisted to `.wrap`, keyframes use `calc()` (`3b76929`)
7. Expanded handsfree overflow → `calc(bars + 54px)` so bars can't push the buttons out (`3b76929`)

`npm run check` and the state smoke test pass. Net result: every pill width and width-keyframe now derives from the same DPI-snapped `--bar-w`/`--bar-gap`, so the visualizer renders evenly and transitions without jumps at any display scale.

PR #158, opened 2026-06-24
fix(injection): skip corrupting clipboard sniff for very_casual

MONKE2525E:
## Problem

Dictating into Chromium-backed rich editors (Quill — used by Discord/Slack) produced scrambled output: correct cleaned text went in, gibberish came out.

Root cause is the Windows contextual-caps **clipboard sniff**. To verify a caret-local "mid-sentence" read before lowercasing, it fires synthetic Shift+Left / Ctrl+C / Right keystrokes immediately before the paste. Those editors report phantom "mid-sentence" text in a visually empty box, so the sniff trips on every dictation, and its keystrokes scramble the paste (the "HellHi.o" corruption the code already warns about).

## Fix

For the `very_casual` profile the sniff is pure downside: both new-sentence and mid-sentence contexts preserve the model's lowercase output, so the sniff can never change the casing decision — but it can still corrupt. Gate it out for `very_casual` so injection just pastes, with no destructive keystrokes.

`casual` / `formal` still sniff (they need the capitalization disambiguation); their behavior is unchanged.

## Validation

- The cleaned text is computed and stored correctly; only the keystroke sniff was garbling delivery. Confirmed against on-device history (raw vs. clean) — the model output was coherent.
- New unit assertion: `very_casual` returns `false` from the sniff gate even on a caret-local mid-sentence read.
- `cargo test injection` 6/6 pass; `cargo clippy -- -D warnings` clean.

## Follow-up (not in this PR)

`casual` / `formal` dictation into Chromium/Quill editors can still hit the same corruption. A complete fix would detect Chromium/contenteditable controls and skip the sniff regardless of profile, or make the sniff non-destructive.

### DIFF
--- a/src-tauri/src/core/injection/mod.rs
+++ b/src-tauri/src/core/injection/mod.rs
@@ -111,29 +111,42 @@ mod tests {
             true,
             ContextProbeSource::CaretLocal,
             SentenceContext::MidSentence,
+            "casual",
         ));
         // A caret read that already says "new sentence" capitalizes — no sniff.
         assert!(!caret_local_needs_sniff_verification(
             true,
             ContextProbeSource::CaretLocal,
             SentenceContext::NewSentence,
+            "casual",
         ));
         // Empty-field and history sources are already trustworthy.
         assert!(!caret_local_needs_sniff_verification(
             true,
             ContextProbeSource::EmptyField,
             SentenceContext::MidSentence,
+            "casual",
         ));
         assert!(!caret_local_needs_sniff_verification(
             true,
             ContextProbeSource::HistoryFallback,
             SentenceContext::MidSentence,
+            "casual",
         ));
         // Contextual caps off: never sniff.
         assert!(!caret_local_needs_sniff_verification(
             false,
             ContextProbeSource::CaretLocal,
             SentenceContext::MidSentence,
+            "casual",
+        ));
+        // very_casual: the sniff can't change the lowercase outcome, so skip its
+        // destructive keystrokes even on a caret-local mid-sentence read.
+        assert!(!caret_local_needs_sniff_verification(
+            true,
+            ContextProbeSource::CaretLocal,
+            SentenceContext::MidSentence,
+            "very_casual",
         ));
     }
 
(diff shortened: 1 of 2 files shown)


MONKE2525E:
@gemini-code-assist Thanks — both comments are the same point (the Windows path ignores `clipboard_sniff_enabled`), so let me explain why I validated it but deliberately did **not** wire it in.

The only value that flows into `clipboard_sniff_enabled` is the `macos_clipboard_sniff_enabled` setting. That setting:

- defaults to `false`, and
- has no UI anywhere in the app (it's a backend-only, macOS-oriented opt-in).

So on Windows it is **always** `false` in practice, with no way for a user to turn it on. Gating the Windows sniff on it would therefore make the sniff dead-by-default for every Windows user — silently reverting the Chromium/Electron phantom-caret capitalization fix the sniff was added for (`casual`/`formal` mis-capitalizing the first letter in a visually empty Chromium box), for everyone, not just `very_casual`. That's a much larger behavior change than a corruption hotfix should make, and it can't be undone from the UI.

So I kept the gate targeted: skip the sniff only for `very_casual`, where it can never change the casing decision (both new- and mid-sentence preserve the model's lowercase) and is therefore pure downside.

Your underlying concern is fair, though — a macOS-named, default-off, UI-less setting shouldn't be the thing that implicitly governs Windows. The right way to make the sniff configurable on Windows is one of:

1. promote it to a real cross-platform toggle (rename, Windows-sensible default, and a UI control), or
2. skip the sniff specifically for Chromium/contenteditable controls regardless of profile — which also removes the corruption for `casual`/`formal`. That one needs the control's class threaded through the injection context probe (it isn't today), so it's a dedicated change rather than a hotfix.

I noted that as the explicit follow-up in the PR description. Does option 2 sound like the direction you'd prefer for the complete fix?

MONKE2525E:
@gemini-code-assist Agreed, and thanks for confirming. I'll keep this PR scoped to the targeted `very_casual` corruption hotfix, and open a dedicated follow-up for Option 2 — threading the control's class through the injection context probe so the sniff is skipped for Chromium/contenteditable controls across all profiles.

If you have any remaining feedback on this PR as it stands, let me know — otherwise I'll treat it as ready.

PR #159, opened 2026-06-24
feat(audio): exclusive microphone access while dictating (macOS)

MONKE2525E:
## Thinking Path

> - Verenu is a Windows/macOS AI dictation app — hotkey hold → audio capture → transcription → LLM cleanup → text injection.
> - Subsystem: audio/system — specifically the microphone-capture path and the existing system-audio mute control in `src-tauri/src/system/volume.rs`, plus the settings + first-run setup UI.
> - Gap: Verenu can already mute system *output* while dictating (`mute_audio`), but nothing stops *other apps* (Discord, Zoom, etc.) from simultaneously capturing the mic while you dictate.
> - Why now: requested feature, and the natural input-side analog of the existing output-mute — useful for anyone dictating mid-call without manually juggling their mic.
> - Feasibility was checked before building: macOS exposes CoreAudio hog mode (`kAudioDevicePropertyHogMode`) for true exclusive device ownership; Windows has **no** supported per-app capture-mute API (WASAPI exclusive mode would require a cpal rewrite and still wouldn't evict an app already on a call), so it is macOS-only by design.
> - This pull request adds an opt-in `exclusive_mic` setting that takes hog mode on recording start and releases it on every stop/cancel path, reusing the CoreAudio plumbing already in `volume.rs`.
> - The benefit is that other apps can't hear you while you dictate, with the toggle hidden on Windows (`{#if isMac}`) and the backend compiled to a no-op there so nothing changes for Windows users.

## What Changed

- `system/volume.rs`: `hog_mic()` / `release_mic()` — acquire/release CoreAudio hog mode on the default input device (mutex-guarded state, warn-once on unavailable, OSStatus-only logging); no-op stubs under `#[cfg(not(target_os = "macos"))]`.
- `data/store.rs` + `commands/settings.rs`: new `exclusive_mic` boolean setting (default **off**) — store constant, `AudioConfig` field + loader, and a `Bool` setting spec.
- `pipeline/session.rs`: acquires exclusive access on recording start (skipped during calibration, matching `mute_audio`); `release_mic` is paired with every existing `unmute` across `commands/recording.rs`, `main.rs` (double-press cancel + Escape), and `pipeline/mod.rs` (both pipeline paths).
- `settings.ts` + `AudioSection.svelte` + `QuickSettingsStep.svelte` / `Setup.svelte`: a macOS-only "Exclusive microphone access" toggle in **Microphone** settings and the setup wizard's **System** card (hidden on Windows via `isMac`).

## Verification

- `npm run check` — 0 errors.
- `npm run lint` — ESLint + `clippy -- -D warnings` clean.
- `npm run test:rust` — 252 passed.
- The macOS hog-mode code is `#[cfg(target_os = "macos")]`-gated, so it is compiled only by the macOS CI leg (it cannot be built on a Windows dev machine; the Windows build validates the no-op stubs and all call sites).
- Manual (requires a Mac): enable the toggle → start another mic consumer (Photo Booth / a call app) → hold the dictation hotkey and confirm the other app loses mic input; release, plus the Escape-cancel and double-press-cancel paths, and confirm the mic is returned. Confirm the toggle is absent on Windows (Microphone settings + setup wizard).

## Risks

- Low-to-moderate, scoped to macOS. Hog mode is device-wide and may briefly reconfigure the device; eviction of an app already streaming the mic is driver-dependent and not guaranteed instantaneous. v1 targets the **default** input device only — a custom-selected mic isn't reserved. No effect on Windows (toggle hidden, backend no-op). No DB migration, no API-key/clipboard/hotkey-timing changes; the new toggle reuses the existing `Toggle` component, so the smoke-test contract is unchanged.

## Model Used

- Anthropic Claude Opus 4.8 (`claude-opus-4-8`) via Claude Code — extended thinking + tool use.

## Checklist

- [ ] Smoke tests pass — not run locally; the new rows are macOS-gated (`{#if isMac}`) so they don't render in the Windows smoke environment, and they reuse the existing `Toggle` component, so the contract is unaffected.
- [ ] Tests added or updated where applicable — none applicable; hog mode requires a real audio device and isn't unit-testable in CI.
- [ ] UI changes include before/after screenshots — UI is macOS-only and couldn't be rendered on the Windows dev machine.
- [x] No API keys, secrets, or personal data in code or logs (failures log OSStatus only)
- [ ] If version bumped: not applicable — no version bump in this PR
- [ ] Relevant documentation updated to reflect changes — n/a for this change
- [x] Risks documented above


MONKE2525E:
@gemini-code-assist I validated the exclusive-mic race you flagged and pushed a fix in `60baf73` for this PR.

What I changed:
- registered a monotonic exclusive-mic session ID on the caller thread before spawning the hog-mode worker
- carried that session ID through the recording lifecycle so stop, cancel, and pipeline completion only release the matching session
- stored the owning session ID alongside the hogged device so an older async release cannot clear a newer session
- made late-starting hog attempts self-abort and release immediately if their session is no longer active
- added matching non-macOS no-op session stubs so the cross-platform call surface stays aligned

What I validated:
- the original implementation could let a quick stop/cancel beat the hog worker and leave a late hog acquisition behind
- the updated flow now invalidates stale work instead of letting ownership drift across sessions
- `cargo check` passes and the Rust test suite passes locally (`260 passed`)

Please take another look and let me know if you still see any gaps in the concurrency handling.

MONKE2525E:
@gemini-code-assist I validated the follow-up leak cases you pointed out and pushed a second fix in `defb99a`.

What changed in this round:
- stop paths now release the exclusive-mic session token even if the recording object is already gone
- the hotkey cancel paths now retrieve the recording and exclusive-mic session token independently so a missing recording cannot drop the release work
- the pipeline/session handoff now always takes the exclusive-mic session token, and early bail-outs release it before returning

What I validated:
- your new findings were valid because a retained exclusive-mic session token could be dropped when `session` was `None`
- the updated flow now releases stale exclusive-mic ownership independently from recording-session presence
- the Rust test suite passes locally again (`260 passed`)

Please review the latest head again and let me know if you still see any gaps.

MONKE2525E:
@gemini-code-assist I validated the latest restart and teardown concerns and pushed another follow-up in `516dc28`.

What changed in this round:
- if a new session starts while the same input device is still hogged by the previous session, ownership is now transferred to the new session instead of silently returning early
- if the default input device changed, the stale hog is cleared before reacquiring
- the pipeline teardown paths now release exclusive mic access only after `session.stop()` has finished
- I applied the same release-after-stop ordering to the direct stop/cancel paths for consistency

What I validated:
- the rapid restart race was real because a new session could miss exclusive access while the old async release was still pending
- the earlier pipeline ordering did release hog mode before the recording stream had fully stopped
- the Rust test suite passes locally again (`260 passed`)

Please review the latest head again and let me know if you still see any gaps.

MONKE2525E:
@gemini-code-assist I validated the latest lock-order and executor-thread concerns and pushed another follow-up in `ec58529`.

What changed in this round:
- `hog_mic` now rechecks the active session ID under the lock before transferring ownership of an already-hogged device
- `hog_mic` also rechecks the active session ID under the lock before committing the post-acquire `HogState` update
- `stop_recording` now offloads the blocking `s.stop()` teardown work off the async executor thread, and keeps the exclusive-mic release tied to that teardown

What I validated:
- the new lock-order concerns were valid because the earlier ownership-handoff path still had a tiny stale-session window under concurrent cancel/restart timing
- the `stop_recording` path was still doing blocking audio teardown directly inside an async command
- the Rust test suite passes locally again (`260 passed`)

Please review the latest head again and let me know if you still see any gaps.

MONKE2525E:
@gemini-code-assist I validated the latest review and pushed another follow-up in `b20d849`.

What I changed:
- `stop_calibration_monitoring` now takes and releases any retained exclusive-mic session instead of dropping the session ID silently
- the recording-start failure path now releases the registered exclusive-mic session ID if `RecordingSession::start` fails after registration

What I validated:
- both of those were real leak paths in the current code
- the Rust test suite passes locally again (`260 passed`)

What I did not change in this round:
- I did not introduce a broad RAII mic guard yet. I validated the specific new leaks Gemini found and fixed those directly, but the current ownership crosses async commands, spawned teardown threads, and pipeline handoff paths. I want to avoid swapping in a wider guard abstraction unless Gemini still sees a concrete remaining correctness gap after these leak fixes.

Please review the latest head again and let me know if you still see any actionable gaps.

MONKE2525E:
@gemini-code-assist I validated the latest review points and pushed a focused follow-up in `9df926f`.

What I changed:
- Only register an exclusive-mic session id when exclusive mic mode is actually active on macOS, so non-macOS builds no longer carry a dummy session token.
- Moved the async stop/release/unmute cleanup paths onto the runtime blocking pool instead of spawning ad hoc OS threads.
- Moved the exclusive-mic hog call onto the blocking pool as well for the same reason.

I reran the Rust test suite after the changes and it stayed green with 260 passing tests.

Please take another look and let me know if you spot anything else.

MONKE2525E:
@gemini-code-assist I validated the new default-vs-custom device concern and pushed a focused follow-up in `40669ee`.

What I changed:
- exclusive mic session registration now only happens when we are recording from the default input device
- if a user selected a named custom microphone, we skip hog-mode acquisition instead of incorrectly hogging the system default mic

What I validated:
- `RecordingSession::start` can record from an explicitly selected device
- `hog_mic` still acquires exclusive access on the default input device only
- without this guard, enabling exclusive mic while using a custom device could leave the active mic unprotected while locking the wrong one
- the Rust test suite passes locally after the change (`260 passed`)

Please review the latest head again and let me know if you see anything else.

PR #157, opened 2026-06-23
refactor(prompts): sharpen cleanup intensity contracts

MONKE2525E:
## Thinking Path

> - Verenu is a Windows and macOS AI dictation app that turns held-hotkey audio into transcribed, cleaned-up text for injection into another app.
> - This change stays inside the cleanup prompt assembly layer in `src-tauri/src/api/prompts`, which defines the instructions sent to the cleanup model after transcription.
> - The current intensity presets were vague and overloaded: `light`, `medium`, and `high` did not clearly separate minimal cleanup, normal cleanup, and aggressive compression.
> - That ambiguity matters now because the roadmap explicitly calls out cleanup quality, prompt contracts, and preserving user detail as release-blocking work.
> - This pull request rewrites the cleanup intensity instructions so each preset has a more explicit contract about what it must preserve, what it may change, and how aggressive it should be.
> - It also updates the prompt tier tests so the suite validates the new wording instead of stale expectations from the previous contract.
> - The benefit is tighter model guidance for cleanup behavior, especially around preserving detail in `medium` and avoiding over-compression in `light`.

## What Changed

- Rewrote `role_line()` and `intensity_rules()` in `src-tauri/src/api/prompts/cleanup_rules.rs` so `light`, `medium`, and `high` each have clearer intent and stronger MUST/MUST NOT constraints.
- Added an explicit `medium` role line instead of falling back to a generic default description.
- Tightened `light` prompts to preserve sentence structure and content while only removing speech noise.
- Tightened `high` prompts to allow aggressive compression while explicitly preserving core meaning and important specifics.
- Replaced the generic fallback cleanup wording with `medium`-specific tier text focused on smoothing speech artifacts without aggressive compression.
- Updated prompt tests in `src-tauri/src/api/prompts/tests.rs` to assert the new tier-specific wording.

## Verification

- `npm run check`
- `npm run lint`
- `npm run test:rust`
- Manual scope verification:
  - `git branch --show-current` -> `fix/cleanup-prompt-intensity-rules`
  - `git merge-base HEAD origin/dev` -> `85c6566fbbf7bb5792481ebd765f17d587a0c5bc`
  - `git log --oneline origin/dev..HEAD` -> `b9273fa refactor(prompts): sharpen cleanup intensity contracts`
- Smoke tests not run because this is a backend prompt-contract change with no UI surface.

## Risks

[...]

PR #160, opened 2026-06-24
Release 0.15.0

MONKE2525E:
Release PR cutting **Verenu 0.15.0** from `dev` into `master`.

## Included since 0.14.1
- **Target-Monitor Dictation Pill** — pill shows on the monitor you're typing on (#154)
- **Sharper Pill Visualizer** — even visualizer bar widths across mixed-DPI monitors (#155)
- **Smarter Contextual Capitalization** — app-mapping style leak fix + caps for empty/Chromium text boxes (#144, #153)
- **Cleaner Casual Injection** — skip a clipboard sniff that could corrupt `very_casual` output (#158)
- **Sharper Cleanup Intensity** — tightened cleanup intensity contracts (#157)
- **Native Update Notifications** — surfaces new versions in-app (#141)
- **macOS Reliability** — permissionless hotkey (RegisterEventHotKey), stable code signing, paste fixes, ad-hoc DMG signing (#139, #140)
- **Leaner Backend** — backend-owned settings store, dependency slimming, hardened logging, module splits (#142, #143)

## Release checklist
- [x] Version bumped to 0.15.0 (package.json, tauri.conf.json, Cargo.toml, Cargo.lock)
- [x] dev reconciled with master
- [ ] PR Checks / CodeQL / Dependency Review green
- [ ] Installers built (Windows local, macOS via CI)
- [ ] VirusTotal links collected
- [ ] Merge + tag `Verenu-0.15.0-beta` + GitHub release

Do not merge until installers are scanned and VirusTotal links are in the release notes.


MONKE2525E:
@gemini-code-assist Thanks for the review. I looked at the one finding individually and want to walk through what I found, because I reached a different conclusion and want your take.

**Finding: encoded-traversal checks (`%2e`/`%2f`/`%5c`) in `is_authorized_release_asset_url` are redundant because `path_segments()` yields already-decoded segments.**

I tested this empirically against the `url` crate version this project pins, rather than relying on the doc wording either way. Parsing:

`https://github.com/o/r/releases/download/v1/..%2f..%5cx%2e%2e`

and collecting `path_segments()` gives a final segment of:

`..%2f..%5cx%2e%2e`

So the segments are **not** percent-decoded — the encoding is preserved. That means the literal-character checks (`s == "."`, `s == ".."`, `s.contains('/')`, `s.contains('\')`) never see a `%2f`/`%5c`/`%2e`, because those stay as encoded literals inside a single segment. The encoded-form checks are the only thing rejecting a crafted URL like `..%2f..%2fattacker/a.exe`, where a downstream server or proxy could later decode `%2f` into a real separator and traverse to a different repo. Removing them would reopen that bypass and would also fail the existing `authorized_url_rejects_bypasses_and_foreign_hosts` test.

**What I did instead of removing the checks:** I kept them and added a regression test (`path_segments_preserve_percent_encoding`) that pins this behavior, so the assumption the security check relies on is locked in and self-documenting. If a future `url` version ever starts decoding segments, that test fails loudly and the encoded checks can be revisited at that point.

If you have evidence that a specific `url` version decodes path segments (or a case my test doesn't cover), I'd genuinely like to see it. Otherwise I believe this one is a false positive and the safer call is to keep the checks. Please take another look at this PR and let me know if you spot anything else.

MONKE2525E:
@gemini-code-assist Thanks for the follow-up.

**ROADMAP version heading (`docs/ROADMAP.md`):** Valid catch — fixed. The `## In Progress` heading now reads `0.15.0` to match this release, and I confirmed the other version numbers in that file (`0.10.0`, `0.11.x`, and the `Shipped in 0.10.0` section) are intentional historical/baseline references, so I left those unchanged.

**Earlier `updater.rs` percent-encoding finding:** I see it was dropped from this pass — agreed, and thanks for reconsidering. For the record, I kept the `%2e/%2f/%5c` checks because they're load-bearing (the `url` crate preserves percent-encoding in `path_segments()`), and I added a regression test (`path_segments_preserve_percent_encoding`) that pins that behavior so the assumption is self-documenting.

Both points are addressed and pushed. Please take another look and let me know if you spot anything else.

MONKE2525E:
@gemini-code-assist Thanks — went through all three. Two applied, one respectfully declined with evidence:

**1. Cut-off accessibility comment (`pipeline/session.rs`):** Fixed. Completed the sentence to "...can return a stale cached `false` for the life of the process." Good catch.

**2. Pill window size in `CLAUDE.md`:** Fixed. Updated the documented initial size from `140x44` to `380x44` to match `PILL_WIDTH_POINTS = 380.0` / `PILL_HEIGHT_POINTS = 44.0` in `pipeline/pill.rs`. The doc was stale.

**3. `Option::is_none_or` → `map_or(true, ...)` (`commands/settings.rs`): declined, because the suggested change would break this project's CI.**

Reasoning, validated rather than assumed:
- This repo's `clippy` gate runs `cargo clippy -- -D warnings`. I temporarily applied your `map_or(true, ...)` suggestion and ran that exact command — it fails with `error: this map_or can be simplified` (`clippy::unnecessary_map_or`, denied via `-D warnings`), and clippy's own fix-it suggests reverting to `is_none_or`. So the two tools disagree, and the project's gate sides with `is_none_or`.
- There's no MSRV concern here: there's no `rust-version` in `Cargo.toml` and no `rust-toolchain` file pinning an older compiler, CI builds on Rust 1.94, and `is_none_or` is already established in the codebase (e.g. `data/dictionary.rs`). `is_none_or` was stabilized in 1.82, well below the toolchain in use.

So keeping `is_none_or` is both consistent with the codebase and required by the lint gate. If you'd still prefer `map_or` for portability, it would need an `#[allow(clippy::unnecessary_map_or)]`, which seemed worse than just keeping the idiomatic form.

All pushed. Please take another look and let me know if anything else stands out.

MONKE2525E:
@gemini-code-assist Appreciate the thorough pass. I checked all four against the actual branch state, and this round looks like it came from a truncated view of a large diff (this PR rolls up 14 commits across many files). None of them reproduce — details so you can confirm:

**1. `retry_transcription_impl` truncated at `pipeline/mod.rs:411` (critical):** Not truncated. The function is complete — it ends with the full `finalize_pipeline_completion(app, state, PipelineCompletionContext { ... }).await`, and the body you proposed as the fix is already exactly what's there. The hunk just cut off at the closing brace in the review view.

**2. `stages.rs` / `state.rs` missing (`pipeline/mod.rs:25`, critical):** Both files are present in the branch (`pipeline/stages.rs` ~22 KB, `pipeline/state.rs` ~2.7 KB) and the modules resolve. They appear to have been omitted from the rendered diff, not the PR.

**3. `validate_char_limit` undefined (`data/db/validation.rs`, high):** It is defined — `pub fn validate_char_limit(field, value, limit)` at `data/db/validation.rs:33` — and used in ~20 call sites across `data/db/dictionary.rs` and `data/db/snippets.rs`.

Independent corroboration for 1–3: the OS-matrix `Rust checks` job compiles the crate on both windows-latest and macos-latest and passes, and `cargo clippy -- -D warnings` is green. Any of those three would be a hard compile error, so CI would be red.

**4. `is_none_or` → `map_or(true, ...)` (`settings.rs:230`, medium):** Same as the prior round — keeping `is_none_or`. Briefly: this repo's gate runs `cargo clippy -- -D warnings`, and I verified that your `map_or(true, ...)` form fails it with `clippy::unnecessary_map_or` (clippy's own fix-it suggests reverting to `is_none_or`). There's no sub-1.82 MSRV (no `rust-version`, no `rust-toolchain` file; CI is on 1.94) and `is_none_or` is already used in `data/db/dictionary.rs`, so it's both required by the lint gate and consistent with the codebase.

So no changes this round. If you're seeing different file contents on your side, it's likely the diff truncation — happy to re-trigger a review if that helps. Otherwise, please take another look and let me know if anything substantive remains.

MONKE2525E:
@gemini-code-assist Good round — one real bug, one false positive.

**1. Clipboard left locked on `GlobalAlloc` failure (`core/injection/windows.rs`): valid, fixed.** You're right: the clipboard was opened first, then `GlobalAlloc(...)?` could return early via `?` without `CloseClipboard`, leaving the system clipboard locked for other processes. I fixed it by handling the allocation failure explicitly — `CloseClipboard().ok()` then return the error — which matches the existing `GlobalLock`/`SetClipboardData` failure paths in that same function. I kept the open→empty→alloc ordering rather than reordering to alloc-before-open, since the minimal change stays consistent with the surrounding cleanup style and is easy to verify. Thanks for catching this.

**2. `set_icon_with_as_template` is not a standard Tauri 2.x API (`main.rs`): declined — it is a real Tauri 2.x API.** It's defined in this project's pinned `tauri` 2.11.1 as `TrayIcon::set_icon_with_as_template(icon, as_template)` (`tauri/src/tray/mod.rs`), which sets the icon and the template flag in one call. The crate compiles cleanly with it — both my local Windows release build and the OS-matrix `Rust checks` job (windows + macos) pass `cargo clippy -- -D warnings`, which wouldn't be possible if the method were missing. Splitting into `set_icon` + `set_icon_as_template` would be equivalent but isn't necessary.

The clipboard fix is pushed. Please take another look and let me know if anything else stands out.

MONKE2525E:
@gemini-code-assist Thanks — I dug into both, and I'm respectfully declining this round, since the concerns are already handled in code that wasn't visible in the diff hunks.

**1. `ensure_value_change_hook` recovery on timeout (`auto_learn/focused_text.rs`):** The current branching is deliberate, and resetting on timeout would actually introduce a bug. The tuple is `(spawned, should_reset_flags)`. Key point about `recv_timeout` semantics here: the hook thread sends its ready signal immediately after `SetWinEventHook`, and the `Sender` is owned by that thread. So:
- `Disconnected` ⇒ the sender was dropped ⇒ the thread exited/panicked ⇒ this is the genuine "thread failed" case, and it already maps to `(false, true)` — it **does** reset and allow a retry.
- `Timeout` can only occur while a sender is still alive ⇒ the thread is still alive (just slow to reach the send). In that state, the thread owns the flags and resets `VALUE_CHANGE_HOOK_SPAWNED` itself on every exit path. If we reset `SPAWNED` here while that thread is still coming up, the next call could spawn a **second** WinEvent hook thread — a duplicate-hook race. So leaving it set on timeout is the safe choice.

In short, the "thread died, flag stuck" scenario you're guarding against is the `Disconnected` arm, which already resets; the `Timeout` arm is "thread alive," where not resetting is correct.

**2. Trim app-mapping fields before saving (`commands/library.rs`):** The runtime lookup is already whitespace- and case-tolerant, so untrimmed storage doesn't cause a silent miss. `resolve_app_mapping` (`pipeline/stages.rs`) matches with `m.exe.trim().eq_ignore_ascii_case(process_name)` and also trims `profile` and `cleanup_intensity` when reading them. So a stored `" notepad.exe "` still resolves against a `notepad.exe` foreground process. Normalizing on save would be a cosmetic tidy of `settings.json`, not a correctness fix, so I left the write path as-is to avoid changing stored data shape.

No changes this round. Please take another look and let me know if anything else stands out.

MONKE2525E:
@gemini-code-assist This is the same point as the previous round, so let me give the definitive evidence — `set_icon_with_as_template` does exist and is not a compilation error.

It's a real method on Tauri 2.x `TrayIcon`, defined in this project's pinned `tauri` **2.11.1**:
- `tauri-2.11.1/src/tray/mod.rs:577` — `pub fn set_icon_with_as_template(&self, icon: Option<Image<'_>>, as_template: bool) -> Result<()>`
- It's also wired through `tauri-2.11.1/src/tray/plugin.rs:207`

It's a convenience that sets the icon and the template flag in a single call — equivalent to `set_icon(...)` followed by `set_icon_as_template(...)`, which is exactly your suggested replacement. Both forms compile; there's no missing API here.

Hard proof that it compiles: this crate is built on every push by the OS-matrix `Rust checks` job on both windows-latest and macos-latest, plus my own local Windows release build, all under `cargo clippy -- -D warnings`. A call to a non-existent method is a hard `E0599` error, so none of those could be green if the method were missing — and they are green.

So I'm keeping the single-call form. Your suggested two-call version is functionally identical, just more verbose, so there's no correctness reason to switch. If your indexer is resolving an older Tauri version than the 2.11.1 this repo pins, that would explain the mismatch. Otherwise, I believe everything raised so far is addressed — please let me know if there's anything new.

PR #163, opened 2026-06-25
feat(audio): pause Windows media sessions during dictation

MONKE2525E:
## Thinking Path

> - Subsystem: audio (recording session lifecycle + Settings UI)
> - Today the only option to stop background audio from bleeding into dictation is "Mute PC Audio," which silences the entire system volume, including audio the user may still want playing quietly in the background.
> - A more targeted alternative is to pause whatever media session is actively playing (e.g. Spotify, YouTube, a podcast app) for the duration of dictation, then resume it afterward, rather than muting everything.
> - This pull request adds a Windows-only "Pause media while dictating" toggle that uses the `GlobalSystemMediaTransportControlsSession` API to pause active playing sessions when recording starts and resume them after dictation finishes (only if they're still paused, so a manual resume by the user during dictation isn't overridden).
> - The benefit is finer-grained audio behavior during dictation: media gets paused/resumed cleanly instead of the whole system going silent.

## What Changed

- Added `src-tauri/src/system/media_control.rs`: Windows-only `GlobalSystemMediaTransportControlsSession`-backed pause/resume of active media sessions, with a generation counter to avoid races between overlapping dictation sessions, plus a `DictationMediaPauseGuard` RAII guard to guarantee resume on early pipeline exit.
- Added the `pause_media_during_dictation` setting (`store.rs`, `settings.rs`, frontend `settings.ts`/`tauri.ts`), defaulting to `false`.
- Wired pause/resume calls into the recording session start (`pipeline/session.rs`), stop paths (`commands/recording.rs`, `main.rs` hotkey handlers), and pipeline completion (`pipeline/mod.rs`).
- Added a "Pause media while dictating" toggle to the Audio settings section, shown only on Windows (`AudioSection.svelte`).
- Added the `Media_Control` feature flag to the `windows` crate dependency in `Cargo.toml`.

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml` - 263 passed, 1 ignored (includes new `media_control` unit tests covering pause/resume state transitions).
- `npm run check` - 0 errors.
- Manual: enabled the toggle, played audio in a media app exposing Windows media controls, dictated - playback paused on hotkey-hold and resumed after injection completed.

## Risks

- Low risk. Feature is off by default and Windows-only (no-op on macOS). Resume is gated on the session still being in the `Paused` state, so it won't fight a user who manually resumed playback mid-dictation.

## Model Used

- Claude Sonnet 4.6 (Anthropic), via Claude Code CLI.

## Checklist

- [x] Risks documented above

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -49,6 +49,7 @@ windows = { version = "0.61", features = [
   "Win32_System_Registry",
   "Win32_Media_Audio",
   "Win32_Media_Audio_Endpoints",
+  "Media_Control",
   "Win32_System_Com",
   "Win32_UI_Accessibility",
   "Win32_Storage_FileSystem",
--- a/src-tauri/src/commands/recording.rs
+++ b/src-tauri/src/commands/recording.rs
@@ -255,6 +255,7 @@ pub async fn stop_recording(
     if let Some(s) = session {
         let _ = s.stop();
         crate::media::sound::coordinated_unmute();
+        crate::system::media_control::end_dictation_media_pause();
     }
     pipeline::hide_pill(&app);
     Ok(())
@@ -267,7 +268,15 @@ pub async fn stop_handless_mode(
 ) -> Result<(), String> {
     crate::core::hotkey::set_handless_active(false);
     crate::core::hotkey::reset_chord_state();
-    lock_state(&state)?.handless = false;
-    tauri::async_runtime::spawn(pipeline::run_pipeline(app, state.inner().clone()));
+    let has_session = {
+        let mut st = lock_state(&state)?;
+        st.handless = false;
+        st.session.is_some()
+    };
+    if has_session {
+        tauri::async_runtime::spawn(pipeline::run_pipeline(app, state.inner().clone()));
+    } else {
+        crate::system::media_control::end_dictation_media_pause();
+    }
     Ok(())
 }
--- a/src-tauri/src/commands/settings.rs
+++ b/src-tauri/src/commands/settings.rs
@@ -119,13 +119,14 @@ const SETTING_SPECS: &[SettingSpec] = &[
     setting_spec(store::APP_MAPPINGS, SettingKind::AppMappings, true, true),
     setting_spec(store::NOISE_REDUCTION, SettingKind::Bool, true, true),
     setting_spec(store::MUTE_AUDIO, SettingKind::Bool, true, true),
-    setting_spec(store::MIC_GAIN, SettingKind::MicGain, true, false),
     setting_spec(
-        store::PLAY_START_STOP_SOUNDS,
+        store::PAUSE_MEDIA_DURING_DICTATION,
         SettingKind::Bool,
         true,
         true,
     ),
+    setting_spec(store::MIC_GAIN, SettingKind::MicGain, true, false),
+    setting_spec(store::PLAY_START_STOP_SOUNDS, SettingKind::Bool, true, true),
     setting_spec(store::SETUP_COMPLETE, SettingKind::Bool, true, false),
     setting_spec(store::APP_CONTEXT_HINT, SettingKind::Bool, true, true),
     setting_spec(store::AUTO_LEARN_ENABLED, SettingKind::Bool, true, true),
@@ -293,6 +294,24 @@ mod setting_key_tests {
         assert!(!is_exportable_setting_key(store::KEY_GROQ));
         assert!(!is_exportable_setting_key(store::KEY_OPENAI));
         assert!(!is_exportable_setting_key(store::KEY_GOOGLE));
+    }
+
+    #[test]
+    fn pause_media_during_dictation_is_boolean_setting() {
+        assert!(is_readable_setting_key(store::PAUSE_MEDIA_DURING_DICTATION));
+        assert!(is_exportable_setting_key(
+            store::PAUSE_MEDIA_DURING_DICTATION
+        ));
+        assert!(validate_setting(
+            store::PAUSE_MEDIA_DURING_DICTATION,
+            &serde_json::json!(true)
+        )
+        .is_ok());
+        assert!(validate_setting(
+            store::PAUSE_MEDIA_DURING_DICTATION,
+            &serde_json::json!("yes")
+        )
+        .is_err());
     }
 }
 // ---------- API keys ----------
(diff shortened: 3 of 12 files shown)


MONKE2525E:
@gemini-code-assist Thanks for the review — addressed both findings:

1. **Critical (COM init):** Confirmed — `begin()`/`end()` in `media_control.rs` dispatch to `tauri::async_runtime::spawn_blocking`, which hands the closure a fresh thread-pool thread with no COM apartment initialized. The WinRT `GlobalSystemMediaTransportControlsSessionManager` calls would fail with `CO_E_NOTINITIALIZED`. Added a `ComGuard` RAII type (`CoInitializeEx`/`CoUninitialize`) initialized as MTA at the top of both closures, since these are WinRT calls on background threads rather than UI Automation calls (which use STA elsewhere in this codebase).
2. **Medium (double lock):** Confirmed — the handsfree `Cancel` branch was locking `state_hk` twice in a row: once to read `session.is_some()`, then again just to flip `handless = false`. Combined into a single lock acquisition that does both.

Both fixes are in the latest commit, build and `cargo test` pass locally. Could you take another look?

MONKE2525E:
@gemini-code-assist Addressed this follow-up round too — all 4 medium findings were the same root cause: the `Ok`/`Err(poisoned)` match arms in `begin()`, `end()`, `is_current_generation()`, and `store_paused_session()` each duplicated their state-update logic instead of acquiring the guard once. Refactored all four to lock once (recovering from poisoning first), then do the actual update/check a single time. Also added the poison-recovery warning log to the two functions that were silently skipping it, for consistency with `begin()`/`end()`.

Pushed, build + `cargo test` still pass (263 passed). Let me know if anything else stands out.

MONKE2525E:
@gemini-code-assist The macOS CI job was failing — `MediaPlaybackState`, `should_pause_for_dictation`, and `should_resume_after_dictation` in media_control.rs are only consumed by the Windows-only platform module and by the unit tests, so on a macOS non-test build neither consumer compiles in, tripping the `deny(dead_code)` clippy gate. Scoped those three items to `#[cfg(any(test, windows))]` so they only exist where something actually uses them.

Pushed, and all CI checks are now green (Windows + macOS Rust checks, frontend, fast profile, CodeQL, dependency audits). Could you take another look?

MONKE2525E:
@gemini-code-assist Good catch — confirmed. `ComGuard::drop()` was unconditionally calling `CoUninitialize`, even on a failed `CoInitializeEx` (e.g. `RPC_E_CHANGED_MODE` if a reused `spawn_blocking` thread already had a different COM apartment from an earlier task). That would have decremented someone else's COM ref count and could break other COM-dependent code sharing that thread.

Fixed by tracking whether init actually succeeded and only uninitializing in that case — same pattern the existing `ComGuard` in auto_learn already uses, which I should have followed from the start.

Pushed, build/tests/clippy all pass locally. Please take another look when you get a chance.

MONKE2525E:
@gemini-code-assist Good catch, confirmed — this was a real bug. On an immediate cancel, the just-paused session's reported playback status can lag behind the actual pause, so the immediate-restore path's staleness check (meant for the normal end-of-dictation resume) was rejecting the read and silently leaving the session paused forever.

Renamed `restore_session_if_still_paused` to `restore_session(session, force)`: the immediate-restore call now passes `force: true` (we just paused it ourselves, no need to re-check), while the normal end-of-dictation `restore_sessions()` path still passes `force: false` so a user's manual resume mid-dictation isn't overridden.

Pushed, build/tests/clippy all pass locally. Let me know if there's anything else.

MONKE2525E:
@gemini-code-assist please re-review this PR in full, looking at all changed files end to end, not just the most recent diff.

MONKE2525E:
@gemini-code-assist Confirmed, good find — this was a real gap beyond the `force` fix from the previous round. That fix only covered the race *inside* `begin()`'s own pause-enumeration loop. `end()`/`restore_sessions()` (force=false) can still run very soon after `begin()` for a dictation cancelled quickly via hotkey, and WinRT's reported playback status can lag behind an already-acknowledged pause command at that point — silently skipping the resume and leaving media paused indefinitely.

Applied your suggested relaxation: resume unless the session is `Stopped` or `Closed`. Calling `TryPlayAsync` on a session that's already `Playing` (e.g. user manually resumed it mid-dictation) is a harmless no-op, so this doesn't reintroduce the "fighting the user" problem. Updated both unit tests you flagged to match the new semantics.

Pushed, build/tests/clippy all pass locally. Please do another full re-review when you get a chance.

MONKE2525E:
@gemini-code-assist Confirmed, fixed — switched `sessions.Size().unwrap_or(0)` to `sessions.Size()?`. Silently defaulting to 0 meant a real WinRT failure would just fail the pause request with nothing in the logs, since it never reached the existing `log::warn!` at the `begin()` call site. Propagating it now lets that warning actually fire.

Pushed, build/tests/clippy pass locally. Please take another look.

MONKE2525E:
@gemini-code-assist Confirmed, this is a real one — a separate race from the others. A `TryPauseAsync` confirmation issued under an old generation can resolve *after* a newer dictation has already started. The old exact-generation check in `store_paused_session` would reject that late confirmation and immediately force-resume the media, even though the new, still-running dictation needs it to stay paused.

Added the `is_active` flag exactly as suggested: set on `begin()`/`end()`, and `store_paused_session` now hands a late-arriving session off to whichever dictation is currently active instead of requiring an exact generation match.

Pushed, build/tests/clippy pass locally. Please take another look.

MONKE2525E:
@gemini-code-assist I checked this one carefully and don't think it holds up — `HSTRING::to_string_lossy()` does exist for this project's pinned dependency version. It's an inherent method defined directly in the `windows-strings` crate (the one that provides `HSTRING` to `windows`/`windows-core`), not something that would need an extra trait in scope:

```rust
// windows-strings-0.5.1/src/hstring.rs
impl HSTRING {
    pub fn to_string_lossy(&self) -> String {
        String::from_utf16_lossy(self)
    }
    ...
}
```

This matches `Cargo.lock`'s pinned `windows = "0.61.3"` → `windows-core 0.61.2` → `windows-strings 0.5.1`. It also already compiled clean locally (`cargo check`, `cargo clippy -- -D warnings`, `cargo test`) and the `Rust checks (windows-latest)` CI job has passed on every commit in this PR, including the latest one, with this exact line unchanged since the first commit.

Leaving this one as-is since it would just be swapping a correct, lossy-safe method for an equivalent one with no behavior change. Let me know if you were thinking of a different type or crate version.

PR #164, opened 2026-06-26
fix(audio): tighten handsfree sound delay and animate pill move

MONKE2525E:
## Thinking Path

> - Audio/UI subsystem: the synthesized start/stop/error/cancel sound cues in `media/sound.rs`, and the floating dictation pill's per-monitor placement in `pipeline/pill.rs` / `pipeline/pill_position.rs`
> - Two related but distinct timing problems surfaced from real usage: (1) the handsfree start chime was delayed 700ms so it would land after the pill's entry animation, but the pill actually finishes expanding around 350ms after the state event fires, so the chime trailed the pill by several hundred ms; (2) on a two-monitor setup with different Windows display scales (e.g. 100% and 125%), the pill's native window must resize (it's sized in physical pixels per-monitor), and that resize-while-visible made WebView2 visibly stutter recreating its swap chain when crossing monitors
> - Both needed correcting because they make dictation feedback feel laggy/janky exactly at the moment a user is watching the pill for confirmation that it's listening
> - This pull request lowers `START_CUE_HANDSFREE_DELAY_MS` from 700ms to 220ms (tuned by ear), and replaces the pill's instant cross-monitor jump (previously masked by a blocking 120ms blank delay) with a real animated `SetWindowPos` tween
> - The benefit is the handsfree chime now lands close to when the pill visually appears, and switching monitors mid-dictation reads as a deliberate glide instead of a sudden snap-and-flicker

## What Changed

- Lowered `START_CUE_HANDSFREE_DELAY_MS` in `src-tauri/src/media/sound.rs` from `700` to `220`. No other sound-cue logic changed; the normal (non-handsfree) start-cue delay, the "Dictation sounds" toggle gating, and all other cues are untouched.
- Added `src-tauri/src/pipeline/pill_animation.rs`: a hand-rolled, ease-out (no overshoot) multi-step `SetWindowPos` tween (~180ms, ~15 steps) that animates the pill's position+size together when a cross-monitor move requires a real resize. Guarded by a generation counter (`PILL_TWEEN_GEN`) mirroring `sound.rs`'s existing `START_CUE_GEN` pattern, so an in-flight tween can't fight a newer move over the window.
- `pipeline/pill.rs`: `show_pill_msg` now branches on Windows — same-monitor redictation stays exactly as fast/synchronous as before; a real cross-DPI move spawns the tween and defers the reveal (click-through, show, state/error emit, extracted into a shared `reveal_pill`) until it lands. Removed the old blocking `MONITOR_MOVE_SETTLE_MS` sleep-then-jump, since the tween's own duration is now the settle time.
- `pipeline/pill_position.rs`: added a `current_placement` accessor and promoted `dimension_changed` to module visibility so `pill.rs` can detect when a move actually needs animating.

## Verification

- `cargo build` / `cargo test --manifest-path src-tauri/Cargo.toml` - 270 passed, 1 ignored, 0 failed (5 new tests covering the tween's easing curve, exact landing on target, no-overshoot monotonicity in both directions, frame count, and generation-counter mechanics)
- `cargo clippy --manifest-path src-tauri/Cargo.toml` - clean
- `cargo fmt -- --check` - clean on all touched files
- Manual (sound delay): ran `npm run tauri dev`, triggered handsfree dictation repeatedly, iterated the delay down (700ms -> 380ms -> 220ms) by ear against the pill's visual expand animation
- Manual (pill animation): confirmed on the reporter's real two-monitor mixed-DPI setup (1080p @ 100%, 1440p @ 125%) — dictating back and forth between monitors now shows a smooth glide instead of a flicker/jump

## Risks

- Sound delay: low risk, single constant value change, no new code paths.
- Pill animation: touches a delicate, previously-flicker-prone code path (this is the second fix to this exact symptom this cycle). Mitigated by: keeping the synchronous same-monitor path byte-for-byte unchanged in cost, gating the entire animated path behind `#[cfg(target_os = "windows")]` (macOS keeps its existing untouched code path), and a generation counter to prevent overlapping tweens from fighting over the window. The interpolation math is unit tested; the actual native `SetWindowPos`/WebView2 behavior can't be unit tested and was verified manually instead.

## Model Used

- Claude Sonnet 4.6 (Anthropic), via Claude Code (agentic file edits + Bash test/build runs; the pill-animation design went through Claude Code's plan mode with a dedicated Plan-agent design pass before implementation).

## Checklist

- [x] `npm run check` passes (TypeScript) - n/a, no frontend files changed
- [x] `npm run lint` passes (ESLint + Clippy) - clippy clean; no frontend files changed
- [ ] Smoke tests pass (see `Agent-Skills/SmokeTest.md` for commands) - not run, no frontend/UI surface touched
- [x] Tests added or updated where applicable - 5 new unit tests for the tween math in `pill_animation.rs`
- [ ] UI changes include before/after screenshots - n/a, native window motion isn't screenshot-able; verified live instead
- [x] If version bumped: all three files updated together - n/a, no version bump
- [x] Relevant documentation updated to reflect changes - n/a, no doc-facing behavior change
- [x] Risks documented above

### DIFF
--- a/src-tauri/src/media/sound.rs
+++ b/src-tauri/src/media/sound.rs
@@ -43,7 +43,7 @@ const AMPLITUDE: f32 = 0.30; // gentle, fixed master level
 pub const START_CUE_NORMAL_DELAY_MS: u64 = 260;
 /// Delay before the start cue when entering handsfree, so it lands after the
 /// pill's entry animation rather than before it looks like it's listening.
-pub const START_CUE_HANDSFREE_DELAY_MS: u64 = 700;
+pub const START_CUE_HANDSFREE_DELAY_MS: u64 = 220;
 
 /// Generation counter for the pending start cue. Scheduling a new start cue or
 /// cancelling supersedes any still-pending one, so the discarded first tap of a
--- a/src-tauri/src/pipeline/mod.rs
+++ b/src-tauri/src/pipeline/mod.rs
@@ -19,6 +19,7 @@ mod finalize;
 mod fixture;
 mod gates;
 mod pill;
+mod pill_animation;
 mod pill_position;
 mod session;
 mod stages;
--- a/src-tauri/src/pipeline/pill.rs
+++ b/src-tauri/src/pipeline/pill.rs
@@ -5,7 +5,7 @@
 
 use super::SharedState;
 use crate::pipeline::pill_position::PillPlacement;
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use tauri::{AppHandle, Emitter, Manager, Runtime, WebviewWindow};
 
 const PILL_WIDTH_POINTS: f64 = 380.0;
@@ -22,6 +22,12 @@ const PILL_HEIGHT_POINTS: f64 = 44.0;
 /// Every `show_pill_msg` call claims a new generation; a deferred reveal
 /// only runs if its generation is still current.
 static REVEAL_GEN: AtomicU64 = AtomicU64::new(0);
+/// Tracks whether the pill is currently showing a non-idle frontend state.
+/// The native window itself stays visible even in idle so WebView2 doesn't
+/// suspend, which makes `pill.is_visible()` a bad proxy for "the user can
+/// already see the pill." We only animate when a prior non-idle state was
+/// actually on screen; otherwise the first visible reveal should be instant.
+static PILL_VISUALLY_ACTIVE: AtomicBool = AtomicBool::new(false);
 
 fn create_pill_if_needed(app: &AppHandle) {
     if app.get_webview_window("pill").is_some() {
@@ -52,11 +58,15 @@ pub(crate) fn show_pill(app: &AppHandle, state: &str) {
 /// monitor it never needs to resize or reposition after its first
 /// appearance. Handsfree's click-capture zone is therefore wider than the
 /// visible pill; clicks in the empty space around it are swallowed instead
-/// of passing through while handsfree is active. Moving to a monitor with a
-/// different scale factor still resizes it — on Windows that resize is
-/// animated (see `pill_animation.rs`) instead of jumping instantly, since an
-/// instant cross-DPI resize on this always-visible window made WebView2
-/// visibly stutter recreating its swap chain.
+/// of passing through while handsfree is active. Moving to a different
+/// monitor, whether or not its scale factor differs, animates the move on
+/// Windows (see `pill_animation.rs`) instead of jumping instantly, since an
+/// instant cross-monitor move on this always-visible window either visibly
+/// snapped (same-DPI repositions) or made WebView2 stutter recreating its
+/// swap chain (cross-DPI resizes). Only animates if the pill was already
+/// visible - the very first reveal (or one after a long idle period where
+/// its cached geometry might be stale) should never be held back by an
+/// animation nobody can see yet.
 fn show_pill_msg(app: &AppHandle, state: &str, message: Option<&str>) {
     create_pill_if_needed(app);
     let Some(pill) = app.get_webview_window("pill") else {
@@ -74,12 +84,16 @@ fn show_pill_msg(app: &AppHandle, state: &str, message: Option<&str>) {
 
     #[cfg(target_os = "windows")]
     if let Some(current) = super::pill_position::current_placement(&pill) {
-        let needs_animated_move =
-            super::pill_position::dimension_changed(current.width as f64, placement.width as f64)
-                || super::pill_position::dimension_changed(
-                    current.height as f64,
-                    placement.height as f64,
-                );
+        let visually_active = PILL_VISUALLY_ACTIVE.load(Ordering::SeqCst);
+        let needs_animated_move = visually_active
+            && (super::pill_position::dimension_changed(
+                current.width as f64,
+                placement.width as f64,
+            ) || super::pill_position::dimension_changed(
+                current.height as f64,
+                placement.height as f64,
+            ) || super::pill_position::position_changed(current.x, placement.x)
+                || super::pill_position::position_changed(current.y, placement.y));
 
         if needs_animated_move {
             let app = app.clone();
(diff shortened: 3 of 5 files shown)


MONKE2525E:
@gemini-code-assist please review

MONKE2525E:
@gemini-code-assist Thanks for the review — I checked all three findings and they're all real, so I fixed all three (commit 28c92a3):

1. **current_placement falling back to (0,0,0,0)**: confirmed — if reading the window's current size/position fails (or returns zero-sized, which can happen very early in init), the old code silently used `(0, 0)` as the fallback. That fake value would then feed into the animated tween as the "from" geometry, so the pill would visibly grow in from the screen's top-left corner at 0x0 instead of skipping the animation. Changed `current_placement` to return `Option<PillPlacement>`, returning `None` on read failure or a zero-sized result.

2. **Caller needs to handle the new Option**: updated accordingly — the animated-move branch in `show_pill_msg` now only runs when `current_placement` returns `Some`. When it's `None`, execution falls through to the existing synchronous/instant placement path, so the pill still gets positioned correctly, just without the animation for that one call.

3. **Tween loop should bail out if the window's HWND can't be read mid-animation**: confirmed — previously, if the pill window were destroyed while a tween was in flight, the loop would just skip that frame's `SetWindowPos` call but keep sleeping through every remaining frame before exiting. Changed it to return immediately the first time `pill.hwnd()` fails, so the background thread doesn't idle through a dead animation.

All three were genuine edge cases (window not fully initialized yet, or destroyed mid-tween) rather than something I disagreed with, so no pushback here — happy to hear if you see anything else.

MONKE2525E:
@gemini-code-assist please review

MONKE2525E:
Addressed the new finding from the latest review pass (commit a7e2f25):

**Raw OS thread vs Tauri's async runtime**: confirmed and fixed — `animate_pill_placement` was using `std::thread::spawn` + `std::thread::sleep` for the tween loop, which is inconsistent with how the rest of the codebase handles this exact pattern. `media::sound::play_start_delayed_then` already implements the same shape (a generation-counter-guarded delayed action) using `tauri::async_runtime::spawn` + `tokio::time::sleep`. Switched the tween driver to match that convention, avoiding the OS thread-creation overhead for what's a short-lived UI animation.

Verified: build, full test suite (270 passed), and clippy with `-D warnings` (matching CI's exact invocation) all clean after the change.

MONKE2525E:
@gemini-code-assist please review

MONKE2525E:
Addressed the high-priority finding from the latest review (commit 3ff4788):

**Stale-state overwrite race condition**: confirmed and fixed — I traced through the actual sequence and it's real. The animated cross-monitor path defers `reveal_pill` until its ~180ms tween lands. If the dictation state moved on (e.g. recording -> processing, or `hide_pill`'s idle) before that tween finished, the newer call already revealed the correct state immediately (since `next_pill_placement` returns `None` once the placement is no longer stale) — but the earlier tween's `on_complete` would still fire afterward and re-emit its own stale captured state, reverting the pill back to looking like it's still recording/processing.

Added a `REVEAL_GEN` generation counter in `pill.rs`: every `show_pill_msg` call claims a new generation, and the animated path's deferred closure bails out instead of revealing if a newer call has already superseded it. Also updated `hide_pill` to bump the counter and cancel any in-flight tween, so hiding the pill can't be reverted by a tween that was already mid-flight when hide was requested.

Verified: build, full test suite (270 passed), and clippy with `-D warnings` all clean.

MONKE2525E:
@gemini-code-assist please review

MONKE2525E:
Went through the four findings from the latest review individually (commit 1ba3abf):

**Adopted:**
1. **Underscore-prefixed variables actually being read** (`_generation`, `_app`): valid idiom concern. Renamed both to plain names, with an explicit `let _ = x;` on the one platform that doesn't read each — so the variable is never underscore-prefixed on the platform where it's genuinely used, while staying silent (no unused-variable warning) on the platform where it isn't.
2. **`abs_diff` for `position_changed`**: valid, adopted as-is — strictly safer (no theoretical signed-overflow path) with identical behavior for any real screen coordinate.
3. **Skip the tween's final sleep**: valid, adopted via a peekable iterator — there's nothing left to wait for once the last frame lands, so this shaves one redundant ~12ms step off the reveal latency.

**Declined (partially):**
4. **Caching `pill.hwnd()` once before the loop**: I checked this one against Tauri 2.11.1's actual source rather than taking it at face value, since it's in tension with a fix from an earlier round of this same review. `hwnd()` (`window/mod.rs`) calls `dispatcher.window_handle()` — a live lookup, not a value cached at window creation — and can fail once the window leaves Tauri's internal registry. Caching it once before the loop would silently undo the previous fix where a destroyed window mid-tween gets caught and the loop bails immediately instead of sleeping through the remaining frames for nothing. Kept the per-frame re-check for that reason, while still adopting the sleep-skip half of this same suggestion (the two parts are independent).

Verified: build, full test suite (270 passed), and clippy with `-D warnings` (matching CI's exact invocation) all clean.

MONKE2525E:
@gemini-code-assist please review

MONKE2525E:
Validated the latest Gemini review threads individually and pushed the follow-up in commit `0ea7037`.

What I changed:
- Adopted the same-DPI cross-monitor glide fix by including position deltas in the animation trigger, not just size deltas.
- Promoted `position_changed` so that trigger can be reused cleanly.
- Adopted `SWP_ASYNCWINDOWPOS` on the tween path so the Tokio worker is not synchronously blocked by the pill window's UI thread.

One place I intentionally did **not** copy the suggestion literally:
- I did **not** use `pill.is_visible()` as the animation gate. In this codebase the native pill window stays visible even in idle so WebView2 does not suspend, which means `is_visible()` is a bullshit proxy for whether the user can actually see an active pill state. Using it directly would still delay some first visible reveals by ~180ms.
- Instead, I added a small backend-tracked visual-active flag. The move animation now only runs when a non-idle pill state was already on screen, which preserves the intended same-DPI glide without reintroducing hidden-state reveal delay.

Validation:
- `cargo test --manifest-path src-tauri/Cargo.toml` passed: 270 passed, 0 failed, 1 ignored.
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets -- -D warnings` is currently failing on unrelated pre-existing lint debt in tests/audio code, so I did not treat that as a blocker for this pill fix.

@gemini-code-assist please review the latest commit, especially the decision to use the visual-active guard instead of `is_visible()` for the first-visible animation gate.

PR #168, opened 2026-06-28
Use native OS connectivity checks instead of an HTTP probe

MONKE2525E:
## Thinking Path

> - Subsystem: system commands (`commands/system.rs`) - the `check_connectivity` Tauri command the frontend polls every 60s to show the "No internet connection" banner
> - The existing check did a `HEAD` request to `github.com` every poll. Measuring it showed it actually downloads ~3-4KB of response headers (full CSP policy + 3 `Set-Cookie` headers) despite an empty body - real, recurring data usage for something that's supposed to "load nothing"
> - It needs addressing now because the project explicitly targets minimal background data/resource usage (200MB idle RAM target, privacy-conscious networking already noted in the surrounding code comment), and there's a zero-cost alternative available on both target OSes
> - This pull request replaces the active HTTP probe with each OS's native connectivity signal - `INetworkListManager::IsConnectedToInternet()` (COM) on Windows, `SCNetworkReachability` on macOS - falling back to the original `github.com` probe only when the native check is unavailable or inconclusive
> - The benefit is the common case (Windows and macOS, i.e. every shipped build) now costs zero network bytes for a routine connectivity poll, with no behavior change visible to the user and no regression for the rare fallback path

## What Changed

- Added `src-tauri/src/system/connectivity.rs` with `check_native() -> Option<bool>`: Windows reads cached internet state via `INetworkListManager` (COM, zero network traffic - the same signal behind the taskbar Wi-Fi icon); macOS checks local route reachability via `SCNetworkReachability` against the zero address (zero network traffic, no DNS lookup); other platforms return `None`.
- `check_connectivity` in `commands/system.rs` now tries the native check first and only falls back to the existing `github.com` `HEAD` probe when native is unavailable or fails.
- Windows COM call runs inside `spawn_blocking` since COM requires apartment initialization on the calling thread (mirrors the existing `ComGuard` pattern in `system/media_control.rs`).
- Added the `Win32_Networking_NetworkListManager` feature to the `windows` crate dependency in `Cargo.toml`.

## Verification

- `cargo check`, `cargo clippy --manifest-path src-tauri/Cargo.toml -- -D warnings`, and `cargo test` (272 passed, 0 failed) all green on Windows, the platform this sandbox can actually compile.
- `npm run check` (svelte-check, 0 errors) and `npm run lint` (ESLint + Clippy) pass - no frontend changes, contract (`check_connectivity` still returns a plain `bool`) is unchanged.
- Measured before/after header size with `curl -I`: `github.com` HEAD ≈ 3-4KB of headers vs. 0 bytes for the new native path.
- Not run locally: Playwright smoke tests (no frontend/UI files touched by this PR, so skipped) and a live online/offline toggle test on a real Windows machine. Recommend a reviewer toggle Wi-Fi off/on and confirm the "No internet connection" banner still appears/clears within ~60s.
- macOS path is not compile-verified in this environment (no macOS target available here) - it's written against stable, documented `SCNetworkReachability` APIs following the existing minimal-FFI style already used in `core/hotkey/mac.rs`, and will get its first real compile check from the OS-matrix CI job.

## Risks

- macOS `SCNetworkReachability` only confirms a local route exists, not verified internet (e.g. a captive-portal Wi-Fi network with no real WAN would still read as "online"). This is a deliberate accuracy-for-zero-data tradeoff, documented in the code comment in `connectivity.rs`. Apple doesn't expose a third-party-accessible verified-internet signal equivalent to Windows' NCSI.
- No automated test covers the native OS calls themselves (COM/FFI aren't practically unit-testable), consistent with how other native platform code in this codebase (`system/volume.rs`, `system/media_control.rs`) is tested only at the surrounding-logic level, not at the raw syscall level.
- Otherwise low risk: the fallback path preserves the exact prior behavior (same URL, same timeout, same `is_ok()` semantics) when the native check returns `None`.

## Model Used

- Anthropic Claude Sonnet 4.6, via Claude Code (standard tool-use mode, no extended thinking).

## Checklist

- [ ] Smoke tests pass (not run - no frontend/UI files changed in this PR)
- [ ] Tests added or updated where applicable (native OS calls aren't practically unit-testable; see Risks)
- [ ] UI changes include before/after screenshots (N/A - no UI changes)
- [ ] If version bumped: all three files updated together (N/A - no version bump)
- [x] Relevant documentation updated to reflect changes (no existing docs reference the old probe)
- [x] Risks documented above

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -54,6 +54,7 @@ windows = { version = "0.61", features = [
   "Win32_UI_Accessibility",
   "Win32_Storage_FileSystem",
   "Win32_Security_Credentials",
+  "Win32_Networking_NetworkListManager",
 ] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
--- a/src-tauri/src/commands/system.rs
+++ b/src-tauri/src/commands/system.rs
@@ -273,6 +273,14 @@ fn run_launchctl(args: &[&str]) -> Result<(), String> {
 
 #[tauri::command]
 pub async fn check_connectivity() -> bool {
+    // Prefer the OS's own network state (see system/connectivity.rs) — on
+    // Windows this is a local COM call with zero network traffic; on macOS
+    // it's a local routing-table check. Only fall back to an HTTP probe when
+    // the native check is unavailable or inconclusive.
+    if let Some(connected) = native_connectivity_check().await {
+        return connected;
+    }
+
     // Probe github.com (a host Verenu already contacts for release downloads)
     // rather than a third-party beacon like google.com, so the connectivity check
     // doesn't quietly phone a separate domain. Deliberately NOT api.github.com:
@@ -286,6 +294,21 @@ pub async fn check_connectivity() -> bool {
         .send()
         .await
         .is_ok()
+}
+
+#[cfg(windows)]
+async fn native_connectivity_check() -> Option<bool> {
+    // COM requires apartment init on the calling thread, so run on a
+    // dedicated blocking thread rather than whatever tokio worker polls this.
+    tokio::task::spawn_blocking(crate::system::connectivity::check_native)
+        .await
+        .ok()
+        .flatten()
+}
+
+#[cfg(not(windows))]
+async fn native_connectivity_check() -> Option<bool> {
+    crate::system::connectivity::check_native()
 }
 
 // ---------- developer logs ----------
--- a/src-tauri/src/commands/system.rs
+++ b/src-tauri/src/commands/system.rs
@@ -275,10 +275,13 @@ fn run_launchctl(args: &[&str]) -> Result<(), String> {
 pub async fn check_connectivity() -> bool {
     // Prefer the OS's own network state (see system/connectivity.rs) — on
     // Windows this is a local COM call with zero network traffic; on macOS
-    // it's a local routing-table check. Only fall back to an HTTP probe when
-    // the native check is unavailable or inconclusive.
-    if let Some(connected) = native_connectivity_check().await {
-        return connected;
+    // it's a local routing-table check. Only short-circuit on a confirmed
+    // "online" (Some(true)): both NCSI and SCNetworkReachability can report
+    // false negatives behind certain VPNs/enterprise proxies, so a "false" or
+    // unavailable native result still falls through to the HTTP probe rather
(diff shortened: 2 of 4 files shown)


MONKE2525E:
Addressed the review feedback in 9939236.

**What I did:** validated the concern - it's a real, well-documented issue (Windows NCSI and macOS SCNetworkReachability can both report "no internet" behind certain VPNs, enterprise proxies, or a temporarily stuck NCSI state, even when the machine actually has working internet). Applied the suggested fix: `check_connectivity` now only short-circuits on a confirmed `Some(true)` from the native check. A `Some(false)` or `None` both fall through to the existing `github.com` HTTP probe instead of being trusted as "offline" outright. Updated the doc comments in `connectivity.rs` to describe the new contract (only `Some(true)` is trusted; everything else means "ask the active probe").

This keeps the zero-network-traffic benefit for the common case (online, no VPN weirdness) while removing the false-negative risk the review flagged. Verified with `cargo check`, `cargo clippy -- -D warnings`, and `cargo test` (272 passed) after the change.

@gemini-code-assist could you take another look at the updated diff?

MONKE2525E:
@gemini-code-assist /gemini review

(Following up in case the earlier mention didn't trigger a fresh pass — please take another look at the latest commit, which addresses your prior feedback about trusting native connectivity false-negatives.)

MONKE2525E:
Thanks, the `/gemini review` trigger worked - addressed all three in 34208df:

1. **`as_bool()` on `VARIANT_BOOL`** - confirmed it exists (it's in the windows crate's extension impl, just not in the same file as the struct definition) and does exactly what the manual `.0 != 0` was doing. Swapped to it.
2. **Redundant cast** - confirmed `ScNetworkReachabilityRef` and `CfTypeRef` are both `*const c_void` aliases, so the cast was a no-op. Removed it.
3. **Bitwise/comparison precedence** - confirmed Rust's precedence table already makes `flags & FLAG != 0` parse correctly, so this wasn't a bug, but added the explicit parens anyway since it removes any ambiguity for anyone scanning the code with C/C++ precedence rules in mind.

All three validated against the actual crate source before applying (didn't want to take a suggestion on faith if the API didn't exist). Re-verified with `cargo check`, `cargo clippy -- -D warnings`, and `cargo test` (272 passed) after each change.

@gemini-code-assist /gemini review

PR #171, opened 2026-06-28
feat(windows): native title bar with theme-matched caption and hidden icon

MONKE2525E:
## Thinking Path

> - Subsystem: UI / window chrome — the main WebView2 window's title bar
> - The Windows window used a hand-rolled frameless titlebar (custom minimize/close buttons, no maximize, no native snap-layouts), diverging from the standard OS chrome every other app uses and adding code to maintain
> - macOS already runs native decorations at runtime; Windows was the last platform still on the custom titlebar, so aligning it removes that divergence and a dead component
> - This pull request switches the Windows main window to native OS decorations, recolors the native caption to match the app theme, hides the caption icon without breaking the taskbar icon, and enlarges the default window
> - The benefit is a standard, lower-maintenance title bar (native minimize/maximize/close, snap-layouts, system menu) that still matches Verenu's look, with ~150 fewer lines of custom chrome

## What Changed

- Switch the Windows `main` window to native decorations (`decorations: true`, `transparent: false`) via a new `src-tauri/tauri.windows.conf.json` platform override; macOS keeps its existing runtime chrome untouched
- Recolor the native caption to the app's `--paper`/`--ink-mute` theme with `DwmSetWindowAttribute` (`DWMWA_CAPTION_COLOR` / `DWMWA_TEXT_COLOR`) in the Windows branch of `apply_native_main_window_chrome` (`src-tauri/src/app_tray.rs`)
- Hide the caption icon without blanking the taskbar icon: set a transparent `ICON_SMALL` (proper 1-bit all-transparent mask) plus a real `ICON_BIG` via `WM_SETICON` — chosen after `WS_EX_DLGMODALFRAME` proved unreliable here (tao resets the extended style after our call)
- Remove the dead favicon `<link href="/vite.svg">` from `index.html` (the file did not exist; WebView2 was syncing the broken-image glyph as the window icon)
- Enlarge the default window from 1100×720 to 1320×860 (`tauri.conf.json` base + the Windows override)
- Delete the custom `TitleBar.svelte` and its setup-wizard duplicate; drop the now-unused minimize/close wiring from `App.svelte`, `SetupShell.svelte`, and `Setup.svelte`
- Add the `Win32_Graphics_Dwm` feature to the `windows` crate in `src-tauri/Cargo.toml`

## Verification

- `npm run check` — passes (157 files, 0 errors, 0 warnings)
- `npm run lint` — passes (svelte-check + `cargo clippy -D warnings`, clean)
- `npm run test:rust` — passes (272 passed, 0 failed, 1 ignored)
- `npm run test:smoke` — passed earlier this session on the identical frontend diff (titlebar markup is not referenced by the smoke contracts)
- Manual (Windows 11, 125% scaling): launched the app and confirmed via screenshots —
  - **Before:** custom frameless bar, minimize + close only, no maximize.
  - **After:** native minimize / maximize / close render and function; caption color matches the app background (pixel-sampled exact match to `#14110e`); caption icon is fully invisible; taskbar / Alt+Tab icon still shows the orange bar-chart logo; close still hides-to-tray rather than quitting; default window is visibly larger.
  - Caption button spacing is standard Win11 (verified identical to Notepad); it is not tightenable without abandoning native decorations.

(Screenshots were captured during development verification; not embedded here.)

## Risks

- Low-to-moderate, **Windows-only**. The change is platform-scoped (macOS untouched) and additive in the backend.
- The caption icon is hidden via raw `WM_SETICON` with a process-lifetime cached `HICON`; a future tao/Tauri upgrade that changes icon handling could make the caption icon reappear (cosmetic only — no crash, taskbar icon unaffected).
- No migrations, no clipboard/injection/hotkey changes, no API-surface changes, no smoke-test contract changes.
- The larger default size may exceed very small (≤768px tall) screens; the OS clamps to the work area, so it degrades gracefully.

## Model Used

- Anthropic Claude Opus 4.8 (`claude-opus-4-8`), via Claude Code with extended thinking and tool use.

## Checklist

- [ ] Tests added or updated where applicable (N/A — native Win32 chrome + config; not unit-testable in this suite)
- [ ] UI changes include before/after screenshots (captured during dev verification; not embedded in this PR body)
- [x] If version bumped: all three files updated together (N/A — no version bump)
- [ ] Relevant documentation updated to reflect changes (no doc changes required)
- [x] Risks documented above


MONKE2525E:
Thanks for the review — both findings were valid and are fixed in ab5175c.

**`make_hicon` AND mask (high)** — confirmed. The mask was built as one byte per pixel, but `CreateIcon` always expects a packed 1-bit-per-pixel bitmap for the AND mask regardless of the XOR mask's bit depth. Fixed by packing it properly (rows padded to a `WORD` boundary), the same way `make_transparent_hicon` already did it correctly right next to it. Re-verified afterward with a real screenshot that the taskbar icon (which goes through `make_hicon`) still renders cleanly and the caption icon (via `make_transparent_hicon`) is still invisible — no regression, just now actually spec-compliant instead of accidentally looking right.

**Redundant `apply_native_main_window_chrome` calls (medium)** — confirmed for both spots. I went a bit further than the two lines flagged: `setup_tray()`/`apply_runtime_icons()` already call this for *both* macOS and Windows internally, so the macOS calls right beside the ones this PR added were just as redundant — that half predates this PR, but since this PR is already touching those exact lines, I removed both rather than leaving an inconsistent half-fix.

Verification on the fix commit: `npm run check` (0 errors), `cargo clippy` (clean), `npm run test:rust` (272 passed), plus a manual screenshot pass confirming no visual regression.

@gemini-code-assist could you take another look at this PR?

MONKE2525E:
Found your follow-up comments on the previous fix commit — both confirmed and fixed in ce4df36.

**Theme-hardcoded taskbar icon (high)** — confirmed. `cached_caption_icons` hardcoded `IconTheme::Dark` and cached the result once for the whole process via a single `OnceLock`, so switching to Light theme would never update the taskbar/Alt+Tab icon. Now the resolved theme is threaded in from the same `resolve_icon_theme` call already used for the caption color, and dark/light real icons are cached in separate `OnceLock`s (the transparent caption icon stays in one cache — it never depends on theme).

**Un-zeroed XOR bytes for transparent pixels (medium)** — confirmed. `make_hicon` set the AND mask bit but left the original colour bytes in place for masked pixels. Now zeroes all four BGRA bytes wherever the AND mask marks a pixel transparent. For this app's actual icon (background pixels already happen to be `[0,0,0,0]`) it had no visible effect, but it's a real correctness gap for a reusable helper, so it's fixed properly rather than left as a coincidence.

Re-verified after this round: build clean, `npm run check` (0 errors), `cargo clippy` (clean), `npm run test:rust` (272 passed), and a screenshot confirming the taskbar icon still renders correctly (clean orange bars, no corruption from the byte-zeroing change).

@gemini-code-assist another pass whenever you're able — let me know if there's anything else.

MONKE2525E:
Two more confirmed and fixed in 32718b3.

**Blanked title breaking taskbar/Alt+Tab/accessibility (high)** — confirmed and fixed exactly as suggested: the title now stays "Verenu" (from `tauri.conf.json`) and `DWMWA_TEXT_COLOR` matches the caption background instead of the title being blanked. Before committing I specifically checked whether this would also blank the minimize/maximize/close button glyphs, since I had `DWMWA_TEXT_COLOR` set to a contrasting color for them — turns out it doesn't: a screenshot test confirmed the buttons stay clearly visible even with `DWMWA_TEXT_COLOR` set to the background color, while the title text blends away. Windows colors those button glyphs independently based on the caption color's luminance, not `DWMWA_TEXT_COLOR`. So this was a clean improvement with no tradeoff.

**`make_hicon` bounds guard (medium)** — confirmed and added. Worth noting: this repo's conventions generally avoid defensive checks for scenarios that can't currently happen, and right now both call sites do pass a correctly-sized buffer. I added it anyway because it's consistent with this exact function's own existing convention of failing soft instead of panicking (`CreateIcon`'s `Result` already becomes `0` via `unwrap_or(0)` rather than unwrapping) — the size/buffer-length relationship is a caller contract the compiler doesn't enforce, so a cheap early return here is in the same spirit as the existing failure handling, not new defensive scaffolding.

Re-verified end to end: build clean, `npm run check` (0 errors), `cargo clippy` (clean), `npm run test:rust` (272 passed), and screenshots confirming all of: caption icon invisible, title text invisible, buttons still visible and clickable, taskbar icon still correct.

@gemini-code-assist another look whenever you're able.

MONKE2525E:
Checked this one, and it doesn't hold up — explaining why rather than applying it blind.

**`SendMessageW` `Some(...)` wrapping (medium)** — not applied. The suggestion says `WPARAM`/`LPARAM` don't need wrapping in `Some(...)`, but the actual signature of the safe `SendMessageW` wrapper in the pinned `windows` crate (0.61.3) is:

```rust
pub unsafe fn SendMessageW(hwnd: HWND, msg: u32, wparam: Option<WPARAM>, lparam: Option<LPARAM>) -> LRESULT
```

It takes `Option<WPARAM>` / `Option<LPARAM>`, not bare values. I tried applying the suggested change directly to confirm — it fails to compile with `expected Option<WPARAM>, found WPARAM`, and rustc's own suggested fix is to add the `Some(...)` wrapping back. So the current code is already correct as written; no change made here.

Everything else from this pass is unrelated to this PR's scope (no other findings this round), so there's nothing further to fix right now.

@gemini-code-assist if there's anything I'm missing, let me know — otherwise this should be a clean stopping point.

MONKE2525E:
C:/Program Files/Git/gemini review

MONKE2525E:
Thanks for confirming — glad the SendMessageW reasoning checked out, and good to hear the rest holds up. This PR is ready for human review/merge whenever convenient.

MONKE2525E:
@gemini-code-assist could you please perform a full code review of this pull request now that all four prior rounds of feedback have been addressed (3 fixed in commits ab5175c, ce4df36, 32718b3; 1 correctly identified as not applicable, explained in a prior comment)? Please re-scan the current diff in full rather than just replying to the most recent comment.

/gemini review

MONKE2525E:
Validated the latest Gemini pass and handled the two active points individually.

Fixed in `75c41c2`:
- `apply_native_main_window_chrome` is now private. That finding was valid because all external call sites had already been removed, so keeping wider visibility was just unnecessary API surface.

Not applied:
- The suggested `SendMessageW` simplification is still incorrect for this repo's pinned `windows` crate version (`0.61.3`). The safe wrapper here takes `Option<WPARAM>` and `Option<LPARAM>`, so removing `Some(...)` fails to compile with `expected Option<...>, found ...`.

Validation after the change:
- `npm run check` passed
- `npm run test:rust` passed (`272 passed, 1 ignored`)
- `cargo clippy --all-targets -D warnings` still fails, but only on pre-existing test-lint issues outside this PR change

@gemini-code-assist please run another full review pass on the current PR diff and let me know if there is anything else that actually needs changing.

MONKE2525E:
Followed up on the latest pass. There is still no code change to make here because the only new feedback is the same `SendMessageW` point, and it is still not valid for this PR's actual dependency version.

Direct local binding proof from the pinned crate:
- `windows-0.61.3` defines `SendMessageW(..., wparam: Option<WPARAM>, lparam: Option<LPARAM>)`
- an older crate line (`windows-0.54.0`) has the generic `Param` form, which looks like the API shape your suggestion is assuming

So for this repository as currently pinned, removing `Some(...)` is still a compile error, not an improvement.

There are no additional valid findings from this pass beyond the visibility fix already shipped in `75c41c2`.

@gemini-code-assist if you have any other feedback on the current PR diff besides the already-invalid `SendMessageW` suggestion, please call it out. If not, please consider this review complete.

PR #180, opened 2026-07-03
fix(pipeline): strip trailing Whisper hallucination sentences

MONKE2525E:
## Thinking Path

> - Subsystem: pipeline (post-transcription hallucination gate in `src-tauri/src/pipeline/gates.rs`)
> - Whisper occasionally appends a well-known boilerplate phrase (e.g. "Subtitles by the Amara.org community.") to the *end* of an otherwise-correct transcription when the recording has a tail of near-silent/background audio after the user stops talking
> - The existing gate (`is_transcription_hallucination`) only matched when the phrase was at the *start* of the output, so a trailing hallucination riding along with real speech was injected verbatim - confirmed from a user's own dev-verbose logs where a "we don't use amara.org" transcription came back with the credit line appended
> - This pull request adds a suffix-stripping pass that trims a trailing hallucinated sentence while preserving real speech before it
> - The benefit is dictated text no longer gets unrelated Whisper boilerplate silently injected at the end

## What Changed

- Added `strip_hallucinated_suffix()` and a `last_sentence()` sentence-boundary helper to `src-tauri/src/pipeline/gates.rs`, reusing the existing `PATTERNS` list but applied to the trailing sentence instead of only the whole output
- Sentence-boundary detection requires punctuation (`.`/`!`/`?`) to be followed by whitespace/end-of-string (or a newline) to count as a boundary, so a period inside something like "Amara.org" is not mistaken for one
- Wired `strip_hallucinated_suffix()` into both the main and retry transcription paths in `src-tauri/src/pipeline/mod.rs`, right after `normalize_transcription_math_artifacts()`
- The existing whole-output hallucination gate now also treats an empty (fully-stripped) result as a hallucination, so audio that was *only* the boilerplate is still dropped silently as before

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml pipeline::` - 57 passed, including 4 new tests: trailing Amara.org credit is stripped, real speech is left untouched, whole-output hallucination still reduces to empty, and internal periods (e.g. "amara.org" mentioned in real speech) don't trigger a false split

## Risks

- Low risk: pure text post-processing confined to the transcription gate, no behavior change for transcriptions that don't match a known hallucination pattern
- `PATTERNS` list is intentionally the same one already used for the whole-output gate, so no new false-positive surface was introduced

## Model Used

- Claude Sonnet 5 (Claude Code), no extended thinking

## Checklist

- [ ] `npm run check` passes (TypeScript) - not applicable, Rust-only change
- [ ] `npm run lint` passes (ESLint + Clippy) - not run
- [x] `npm run test:rust` passes (ran targeted `cargo test pipeline::`, 57/57)
- [ ] Smoke tests pass - not applicable, no frontend/UI change
- [ ] UI changes include before/after screenshots - not applicable, no UI change
- [ ] If version bumped: all three files updated together - not applicable, no version bump
- [x] Relevant documentation updated to reflect changes - none needed
- [x] Risks documented above

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,17 @@
+[advisories]
+# These advisories affect `quick-xml`, which we only depend on transitively
+# through Tauri's own dependency graph:
+#   - quick-xml 0.37.5 via tauri-winrt-notification -> notify-rust
+#     -> tauri-plugin-notification
+#   - quick-xml 0.39.3 via plist -> tauri-utils -> tauri/tauri-build
+# Both parent crates pin quick-xml below the patched 0.41.0 (`^0.37` and
+# `^0.39.2` respectively), so we cannot bump past them ourselves without
+# vendoring or forking - this requires an upstream Tauri/notify-rust/plist
+# release. Neither path parses untrusted external XML at runtime for Verenu
+# (Windows toast notification metadata and build-time plist reading), so the
+# practical risk is low. Revisit and remove these once `cargo update` can
+# pull in quick-xml >=0.41 on its own.
+ignore = [
+    "RUSTSEC-2026-0194", # quick-xml: quadratic run time on duplicate attribute check
+    "RUSTSEC-2026-0195", # quick-xml: unbounded namespace-declaration allocation (DoS)
+]
--- a/src-tauri/src/pipeline/gates.rs
+++ b/src-tauri/src/pipeline/gates.rs
@@ -60,6 +60,80 @@ pub(super) fn is_transcription_hallucination(text: &str) -> bool {
     PATTERNS.iter().any(|p| t.starts_with(p))
 }
 
+/// Removes a trailing sentence that matches a known Whisper hallucination
+/// pattern, leaving any real speech before it intact.
+///
+/// Whisper sometimes bolts one of these phrases onto the *end* of an
+/// otherwise-correct transcription when the recording has a tail of
+/// near-silent or background audio after the user finishes talking (e.g. the
+/// mic keeps capturing for a moment before the hotkey is released). That
+/// differs from `is_transcription_hallucination`, which only catches the
+/// case where the *entire* output is one of these phrases.
+pub(super) fn strip_hallucinated_suffix(text: &str) -> String {
+    let mut current = text.trim().to_string();
+
+    // Bounded loop: guards against pathological repeated matches rather than
+    // expecting more than one hallucinated sentence in practice.
+    for _ in 0..4 {
+        let sentence = last_sentence(&current);
+        if sentence.is_empty() || !is_transcription_hallucination(sentence) {
+            break;
+        }
+        let cut = current.len() - sentence.len();
+        current.truncate(cut);
+        current = current.trim().to_string();
+    }
+
+    current
+}
+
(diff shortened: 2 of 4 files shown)


MONKE2525E:
Thanks for the review — addressed both points:

**High: false positives from prefix-matching on trailing sentences.** Confirmed this was a real bug: `strip_hallucinated_suffix` reused `is_transcription_hallucination`'s `starts_with` matching for every pattern, including short generic phrases like `"please subscribe"` and `"thank you for watching"`. A genuine trailing sentence such as "Please subscribe to our newsletter." or "Thank you for watching over my dog." would have been silently deleted. Added a new `is_sentence_hallucination` helper (scoped to the single trailing sentence, not the whole output) that keeps prefix matching only for patterns that are unambiguous even as a prefix (credit lines, system-prompt echoes), and requires an exact match on the rest — so real speech starting with those words survives while actual hallucinations still get stripped.

**Medium: `raw.trim().is_empty()` → `raw.is_empty()`.** Applied as suggested in both the main and retry pipeline paths — `strip_hallucinated_suffix` already returns a trimmed string, so the extra `.trim()` was redundant.

**Medium: test coverage for the false-positive case.** Added tests covering both examples from the review (the subscribe/newsletter and thank-you/dog sentences survive untouched), plus a test confirming a bare trailing hallucination sentence is still stripped.

All 59 pipeline tests pass locally after the change. Note: the failing "Dependency audits" CI check is a pre-existing `quick-xml` transitive-dependency advisory unrelated to this change (no dependency edits in this PR) and is out of scope here.

@gemini-code-assist could you take another look and let me know if there's anything else?

MONKE2525E:
Thanks — addressed the follow-up too:

**Medium: avoid per-iteration allocation in the suffix-stripping loop.** Confirmed and applied as suggested. \`current\` is only ever truncated from the end and starts already trimmed, so leading whitespace can never reappear across iterations — using \`trim_end()\` + in-place \`truncate()\` instead of \`trim().to_string()\` is behaviorally identical and avoids a reallocation on every loop iteration.

59/59 pipeline tests still pass. Also noting the deprecation notice in your review footer about the consumer GitHub integration sunsetting on July 17, 2026 — appreciate you still getting to this one.

@gemini-code-assist anything else on this pass?

MONKE2525E:
Addressed the CJK internationalization findings too:

**Medium: sentence-boundary detection missing CJK fullwidth punctuation.** Confirmed and fixed. \`last_sentence()\` only recognized ASCII \`.\`/\`!\`/\`?\` as boundaries, and required trailing whitespace after them. CJK sentences end with fullwidth terminators (\`。\`/\`！\`/\`？\`) and conventionally have no space after them, so a CJK dictation followed by a trailing English Whisper hallucination would never get split correctly, letting the hallucination survive. Added fullwidth terminators as unconditional boundaries (no whitespace check needed, since they're unambiguous).

**Medium: trim trailing fullwidth punctuation before exact-match comparison.** Applied as suggested — extended the trailing-punctuation trim in \`is_sentence_hallucination\` to also strip \`。！？，；：\` for consistency with the above.

Added two tests covering both scenarios (CJK sentence followed by a trailing English hallucination, and an exact-match hallucination ending in fullwidth punctuation). All 61 pipeline tests pass.

@gemini-code-assist let me know if there's anything else.

PR #181, opened 2026-07-03
Add Verenu status API integration: provider outage alerts, website link, contact info

MONKE2525E:
## Thinking Path

> - Subsystem: pipeline error handling + settings/UI + a new integration with `api.verenu.com`, the app's website API
> - We mapped the website's new API endpoints and found `/v1/provider-status`, which exists specifically to tell users whether Groq/OpenAI/Google are having real problems - but nothing in the app used it, so a failed dictation gave no signal on whether it was the user's key/quota or an actual provider outage
> - Verenu also just got official contact addresses, and the app's own privacy docs claimed "no Verenu-owned server in the product path," which this feature makes untrue the moment it ships
> - This pull request polls provider status (5 min) and API health (20 min), recheck immediately on a provider-side pipeline failure, adds a raw-check button in the Developer panel, a Website link in Settings > About, and corrects the privacy docs plus adds the new contact emails
> - The benefit is users get an in-app "is it me or is it them" signal after a failure, scoped only to providers they actually use, with zero dictated content ever leaving the device toward Verenu's own server - and the docs now accurately describe that

## What Changed

- `api/service_status.rs` + `commands/service_status.rs`: fetch and filter `api.verenu.com/v1/provider-status` to the user's selected transcription/cleanup providers, gated on the backend's `showToUsers` flag and excluding `operational`/`unknown` statuses (`unknown` means no machine-readable feed, not an actual issue); plus a plain `v1/health` check and an unfiltered raw-fetch command
- `pipeline/stages.rs` + `state.rs`: emit `verenu:recheck-provider-status` when a transcription or cleanup call fails with a provider-side error (quota, or a retryable timeout/429/5xx), so a real outage surfaces immediately instead of waiting for the next poll
- `serviceStatus.ts`: polls provider status every 5 min and API health every 20 min (health has no UI yet - kept warm for future use), plus listens for the immediate-recheck event
- `Home.svelte` + `ProviderStatusBanner.svelte`: show provider alerts as a banner that takes priority over the update-available banner when both are pending
- Settings > Developer: "Run Check" button (`check_provider_status_raw`) dumps the unfiltered API response for debugging
- Settings > About: new Website row linking to verenu.com
- `docs/SECURITY.md`, `docs/SUPPORT.md`, `README.md`: add official contact emails (`security@`, `support@`, `hello@verenu.com`); confirmed GitHub private vulnerability reporting is disabled for this repo and removed the conditional wording
- `README.md`, `docs/DATA_AND_PRIVACY.md`: correct the "Verenu does not run its own servers" / "no Verenu-owned server in the product path" claims now that `api.verenu.com` is contacted for status/update metadata, and document exactly what those requests do and don't carry
- `CLAUDE.md`: document the new module and its alerting rules under a "Verenu Status API" section
- 6 new Rust unit tests for the alert-filtering logic (`api::service_status::tests`)
- `serviceStatus.ts`: null-guard the provider-status response (a real bug caught by the frozen smoke-test mock, which returns `null` for unmocked commands - see Verification)

## Verification

- `npm run check` - 159 files, 0 errors
- `cargo check` / `cargo build` - clean
- `cargo test` - 287 passed, 1 ignored, 0 failed (includes the 6 new `service_status` tests)
- `npm run lint` (svelte-check + `cargo clippy -D warnings`) - clean
- `npm test` (fast profile, 16 smoke/state/contract tests) - all pass. This caught a real crash: `Home.svelte` assumed `providerStatusAlerts` was always an array, but the frozen `tests/smoke/_tauri-mock.cjs` returns `null` for any command it doesn't know about, which cascaded into 6 additional failures until fixed (last commit)
- Manual: hit `api.verenu.com/v1/provider-status` and `/v1/health` directly with curl against the live schema; toggled the backend's `showToUsers` flag live to confirm the banner only appears for a selected, genuinely non-operational provider and stays silent for `operational`/`unknown`
- Manual: Settings > Developer > "Run Check" returns raw JSON; Settings > About > Website opens verenu.com in the default browser

## Risks

- This is the app's first outbound call to a Verenu-owned server. Mitigated: requests are plain GETs with no body, never carrying dictated audio/text/keys/history, documented explicitly in `docs/DATA_AND_PRIVACY.md`
- The alert gate depends on the backend's `showToUsers` flag; the client independently excludes `operational`/`unknown` as a backstop so a careless backend flag can't turn a healthy status into a false alert (this exact scenario happened during development and is now covered by a unit test)
- Status/health polling runs on a timer regardless of connectivity; failures are caught and logged as warnings only, never surfaced as user-facing errors

## Model Used

Claude Sonnet 5 (Anthropic), via Claude Code. Standard tool-use mode, no extended thinking.

## Checklist

- [x] Smoke tests pass (`npm test`, fast profile, 16/16)
- [ ] UI changes include before/after screenshots (not captured this pass - changes are two new settings rows and a status banner, all covered by the smoke-test DOM assertions above)
- [x] If version bumped: all three files updated together (N/A - no version bump in this PR)
- [x] Risks documented above


MONKE2525E:
@gemini-code-assist Thanks for the review — addressed all 3 findings in 84bcde5.

**1. High: serde deserialization not tolerant of missing/null fields (`service_status.rs`)** — Confirmed valid and fixed as suggested. `severity`, `userMessage`, `detailsUrl` are now `Option<String>` with `#[serde(default)]`, mapped with `.unwrap_or_default()`. The real impact here is worse than just "one entry missing data" — since `Vec<ProviderStatusEntry>` deserializes as a whole, one provider entry missing an expected field would fail parsing for *every* provider in that response, silently blinding the whole status check (it degrades to "no alerts" rather than crashing, since the caller already treats a parse error as a soft failure, but that's still a real loss of signal).

I extended this a bit past your literal suggestion: `showToUsers` is now `Option<bool>` rather than `bool` with `#[serde(default)]`. `#[serde(default)]` alone only kicks in when the key is *missing* — it doesn't help if the backend sends an explicit `"showToUsers": null`, which still fails to deserialize into a plain `bool`. Making it `Option<bool>` + `.unwrap_or(false)` covers both cases symmetrically with the other three fields.

Added 2 regression tests (`entry_deserializes_when_optional_fields_are_missing`, `entry_deserializes_when_optional_fields_are_explicitly_null`) that deserialize raw JSON directly to prove both cases actually work, not just that the types compile.

**2. High: test helper needed updating for the new `Option` fields** — Valid, and doing so was necessary the moment fix #1 landed (the helper wouldn't compile otherwise). Applied.

**3. Medium: use Svelte 5 `$props()` instead of legacy `export let` in `ProviderStatusBanner.svelte`** — Valid and applied to that file. One thing worth noting: the sibling component right next to it in the same folder, `UpdateBanner.svelte`, still uses `export let` — I didn't touch it, since it's a pre-existing file outside this PR's scope and refactoring it wasn't part of what you flagged. Happy to do that as a follow-up if you'd like the whole folder made consistent.

Verification after the fix: `cargo test` (289 passed), `npm run check`, `npm run lint` (clippy `-D warnings`), and the full `npm test` smoke suite (16/16) all pass.

Could you take another look?

/gemini review

MONKE2525E:
@gemini-code-assist ping — could you take another look at the fixes in 84bcde5 for the 3 issues you found earlier?

/gemini review

MONKE2525E:
@gemini-code-assist Thanks for the follow-up review — addressed all 4 findings in 2d958ad.

**1. High: case-sensitive status/provider ID comparisons** — Valid, fixed with `eq_ignore_ascii_case` on both the status checks and the provider-ID match, exactly as suggested. Added 2 tests proving a differently-cased status doesn't falsely alert and a differently-cased provider ID still matches.

**2. Medium: missing `providers` key should default to empty** — Valid, applied `#[serde(default)]` to that field too, with a test confirming `{}` deserializes to an empty list instead of erroring.

**3. Medium: add explicit timeouts to the three HTTP calls** — Partially pushed back on this one. The shared reqwest client (`api/client.rs`) already sets a client-wide 120s timeout (plus a 10s connect timeout), so the specific claim "without a timeout, requests can hang indefinitely" isn't accurate here — there was already a cap. That said, 120s is calibrated for the app's long-running LLM calls, not a lightweight status poll, so I still added a dedicated 10s timeout to all three calls since a hung connection genuinely shouldn't tie up a background poll cycle that long. Net result matches your suggestion, just with the reasoning corrected.

**4. Medium: guard the "Check status" button against an empty `detailsUrl`** — Valid, and it's a direct consequence of the `Option<String>` fix from your first review (that field can legitimately be an empty string now). Wrapped the button in a conditional.

Verification: `cargo test` (292 passed), `npm run check`, `npm run lint`, and the full `npm test` smoke suite (16/16) all pass.

/gemini review

MONKE2525E:
@gemini-code-assist checking in — do you have any further feedback on the fixes in 2d958ad, or is the PR clear from your side at this point?

/gemini review

MONKE2525E:
C:/Program Files/Git/gemini review

MONKE2525E:
@gemini-code-assist Found and fixed 2 more findings in b02db26 that I'd initially missed — my own comment-search filters were anchored to my own comment timestamps and missed two review passes you ran automatically on push (one against 84bcde5, one against 2d958ad). Doing a full unfiltered pull of the PR's history caught them. Sorry for the noise of the last few pings while I was looking in the wrong place.

**Security-medium: validate URL protocol before opening in `ProviderStatusBanner.svelte`** — Confirmed valid and fixed as suggested. `detailsUrl` comes from an external API response, so `openDetails()` now refuses anything that isn't `http://`/`https://` before handing it to the shell plugin's `open()`.

**Medium: `apiHealthy` kept a stale value on a failed check** — Confirmed the bug, applied a slightly different fix than suggested. Your suggestion set it to `false` on failure; I set it to `null` instead, since the field is typed `boolean | null` specifically to represent "not yet known," and a failed check (e.g. the user's own network hiccupping) doesn't actually confirm the API is down — it just means this cycle couldn't tell. Forcing `false` would overclaim.

Two more from the same passes that I looked at but didn't change:

- **"Use `selected_providers.contains(&p.id)` instead of `.iter().any(...)`"** — This suggestion predates the case-insensitivity fix from your other review (which I already applied). `.contains()` only does exact equality, so switching to it now would silently reintroduce the case-sensitivity bug you flagged elsewhere. Keeping `.iter().any(eq_ignore_ascii_case)` since it's the correct form given that requirement.
- **"5s timeout specifically for the health check"** — I used one shared 10s constant across all three status-API calls instead. Didn't see a strong reason to give the health check a tighter budget than the other two lightweight calls, and one named constant is simpler to reason about than three slightly different magic numbers for calls in the same category.

Verification: `cargo test` (292 passed), `npm run check`, `npm run lint`, and the full `npm test` smoke suite (16/16) all pass.

/gemini review

MONKE2525E:
@gemini-code-assist Fixed in 6867fec.

**Medium: duplicate keys in the `{#each}` block could crash the banner** — Confirmed valid. `alert.providerId` had no uniqueness guarantee from the backend response, and Svelte's keyed `{#each}` throws a fatal error on a duplicate key, which would have taken the whole banner down. Applied your suggested fix (provider ID + loop index).

Verification: `npm run lint` and the full `npm test` smoke suite (16/16) pass.

/gemini review

PR #201, opened 2026-08-10
fix(injection): reduce post-paste verification false positives in Chromium

MONKE2525E:
## Thinking Path

> - Verenu is a Windows and macOS AI dictation app: hotkey hold -> audio capture -> transcription -> LLM cleanup -> text injection.
> - This change is in the **injection** subsystem (`core/injection`), specifically the best-effort post-paste verification that runs right after Ctrl+V.
> - The post-paste verification re-reads the target text field via UI Automation to confirm the paste landed. In Chromium/ProseMirror-style editors (ChatGPT, Claude, Slack), the accessibility tree commits a paste to its own state *before* it catches up to the DOM, and that lag is variable — often well over a second.
> - The old code retried for a short fixed window and then hard-errored with "Paste could not be verified" even though the paste had landed — the false positives reported in Chrome ("it says it didn't paste but it actually did paste").
> - It needs to be addressed because a false failure sticks the "paste failed" pill up, makes the user re-dictate, and can even duplicate the text that already landed.
> - This pull request makes verification wait longer (growing gaps), cross-checks the control's full document/value text as a fallback, and only hard-fails when the accessibility tree *freshly* read the caret as genuinely not containing the injected text. A read that stays frozen on the pre-injection tail is treated as "unverifiable, not failed."
> - The benefit is a near-elimination of false "paste failed" reports in Chrome while still catching genuine failures (paste went to the wrong window / nothing landed).

## What Changed

- `src-tauri/src/core/injection/windows.rs`: post-paste verification now retries 6 times with growing gaps (100ms -> 260ms) instead of 4 fixed 100ms waits, giving Chromium's a11y tree time to catch up.
- `src-tauri/src/core/injection/windows.rs`: added a full-text fallback — when the caret-local read can't confirm the paste, read the control's whole ValuePattern/document text and verify the injected text reached the field (handles stale caret-range anchoring).
- `src-tauri/src/core/injection/windows.rs`: failure decision changed — hard error only when a *fresh* caret-local read genuinely doesn't contain the injected text; frozen reads (identical to the pre-injection tail) are logged and treated as best-effort success.
- `src-tauri/src/core/injection/mod.rs`: extracted the pure text-matching helpers (`injected_suffix`, `paste_tail_matches`, `full_text_confirms_paste`) plus the suffix/min-chars constants so they're unit-testable cross-platform.
- `src-tauri/src/core/injection/mod.rs`: added unit tests covering the tail match and full-text confirmation helpers.

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml` — 461 passed (459 base + 2 new), 0 failed.
- `cargo clippy --manifest-path src-tauri/Cargo.toml --bin verenu` — clean.
- Manual scenario: hold hotkey and dictate into the ChatGPT (ProseMirror) prompt textarea. Previously the failing logs showed `post-paste verify attempt=1..4 verified=false` then a hard "Paste could not be verified" error even though the text landed. After this change the same flow either verifies within the longer window / full-text cross-check, or logs the lenient "treating as best-effort success" path instead of erroring.

## Risks

- Low. The tradeoff is that a paste which *genuinely* fails while UIA never observes the change is now reported as best-effort success rather than a hard error. This is acceptable because: (a) the pre-existing guards already confirm a focused, writable text control before the paste, (b) the transcription is saved to History regardless of injection outcome, and (c) the previous behavior caused frequent false positives in Chromium apps.
- Behavior is Windows-only; the macOS injection path has no post-paste verification and is untouched.

## Model Used

- opencode (`opencode-go/deepseek-v4-flash`, deepseek-v4-flash) — code analysis, implementation, and PR authoring.

## Checklist

- [x] `npm run check` passes (TypeScript) — no frontend changes
- [x] `npm run lint` passes (ESLint + Clippy) — Clippy clean for the changed crate
- [ ] UI changes include before/after screenshots — no UI changes
- [ ] If version bumped: all three files updated together (`package.json`, `src-tauri/tauri.conf.json`, `src-tauri/Cargo.toml`) — no version bump
- [x] Risks documented above

### DIFF
--- a/src-tauri/src/core/injection/mod.rs
+++ b/src-tauri/src/core/injection/mod.rs
@@ -100,6 +100,68 @@ enum CursorContextState {
     },
 }
 
+// Post-paste verification text helpers. Pure functions (no UIA/OS access) so
+// they're unit-testable on every platform; the Windows injection path drives
+// the actual reads and owns the retry timing. They exist because Chromium/
+// ProseMirror-style editors (ChatGPT, Claude, Slack, ...) commit a paste to
+// their own state before the accessibility tree catches up — a caret-local
+// read taken too early still sees the pre-paste text, which used to flag a
+// *successful* paste as failed (the false positives reported in Chrome).
+
+// Capped suffix for the cheap tail check — enough to catch "nothing landed" or
+// "wrong window" without being thrown off by trailing whitespace/newline
+// normalization differences between what we sent and what UIA reports back.
+#[cfg_attr(not(windows), allow(dead_code))]
+const PASTE_VERIFY_SUFFIX_CHARS: usize = 20;
+
+// A `contains` cross-check on the control's full text only counts when the
+// fragment is long enough to be distinctive; short dictations rely on the
+// precise ends-with check instead.
+#[cfg_attr(not(windows), allow(dead_code))]
+const PASTE_VERIFY_CONTAINS_MIN_CHARS: usize = 12;
+
+#[cfg_attr(not(windows), allow(dead_code))]
+fn injected_suffix(injected: &str) -> String {
+    let mut chars: Vec<char> = injected
+        .trim_end()
+        .chars()
+        .rev()
+        .take(PASTE_VERIFY_SUFFIX_CHARS)
+        .collect();
+    chars.reverse();
+    chars.into_iter().collect()
+}
+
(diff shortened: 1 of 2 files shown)


MONKE2525E (commented):
GOOD

PR #202, opened 2026-08-10
Add Insights analytics page

MONKE2525E:
## Thinking Path

> - Subsystem: UI (new sidebar page) + data (new aggregation queries over the transcriptions table)
> - The app records rich per-transcription data (words, spoken_words, duration_ms, api_used, created_at) but the only place any of it surfaced was three numbers on Home's stat card - streaks, speed, and cost were invisible
> - Users have no way to see dictation trends, how fast they talk, or roughly what their transcription/cleanup API usage is costing them across providers
> - This pull request adds an "Insights" sidebar page: hero stats (wpm gauge, total words with a period delta, cleanup/dictionary-fix counts), a daily words chart, an hourly activity chart, a GitHub-style streak heatmap, an estimated-cost donut chart broken down per model, and vocabulary stats - backed by one new `get_insights` backend command
> - The benefit is users get a genuine sense of their dictation habits and estimated spend without leaving the app or querying anything themselves

## What Changed

- New `Insights.svelte` page + `src/lib/views/insights/` components: `HeroStats` (wpm gauge, total words, cleanup counts), `DailyChart`, `HourStrip`, `StreakHeatmap`, `CostBreakdown`, `WordStats`, plus shared `AnimatedNumber`, `ChartTooltip`, `DonutChart` (hand-built canvas donut with filleted segments and hover pop/dim), `pricing.ts` (static per-model rate table), `helpers.ts`, `types.ts`
- New `get_insights` Tauri command (`src-tauri/src/data/db/insights.rs`) aggregating totals, streaks, daily/hourly series, per-model API usage, cleanup effect, and stopword-filtered top words from the `transcriptions` table, with local-day bucketing and zero-filled daily series
- Sidebar/nav wiring for the new page (`Sidebar.svelte`, `App.svelte`, `stores.svelte.ts`, `motion.ts`, `icons.ts`)
- Extracted shared button/dropdown primitives into `src/ui.css`, referenced from `DESIGN.md` (new), `AGENTS.md`, `CLAUDE.md`
- Unrelated one-line fix: the notification bell icon on the home page's global message banner was vertically misaligned against multi-line text - now centered and slightly larger

## Verification

- `npm run check` - 0 errors, 0 warnings
- `cargo check` / `cargo test` (`src-tauri`) - 476 passed, 0 failed, including 12 new tests in `data::db::insights::tests` covering zero-fill, streak counting, local-midnight bucketing, stopword filtering, and the empty-DB case
- Manually exercised in `npm run tauri dev`: range switching (7/30/90/all), hover states on all three charts, streak grid with real and "no data" cells, donut hover on every segment, longest-word expand/collapse
- Verified via Playwright against the running dev server: hero tile heights match across the row, gauge number sits inside the arc with no overlap, delta badge corner position, donut ring has zero rendering gaps and each segment visibly pops on hover, streak grid's real cells match their true weekday row (0 mismatches sampled), hourly chart fills its stretched card height

## Risks

- Low risk - additive UI page plus one new read-only backend command; no changes to the hotkey, pipeline, or injection paths
- Cost estimates are computed client-side from a static rate table (`pricing.ts`) and are explicitly labelled "Estimated" in the UI; they are not billed by Verenu and can drift from actual provider pricing over time

## Model Used

- Claude Sonnet 5 (Anthropic), via Claude Code, extended reasoning + tool use (Read/Edit/Write/Bash/Playwright browser verification)

## Checklist

- [ ] `npm run lint` passes (ESLint + Clippy) - `npm run check` and `cargo check`/`cargo test` were run individually; full `npm run lint` (which also runs clippy) was not re-run after the final commit
- [ ] Smoke tests pass (see `Agent-Skills/SmokeTest.md` for commands) - not re-run after the final rebase
- [x] Tests added or updated where applicable (12 new Rust tests for `get_insights`)
- [x] UI changes include before/after screenshots - see PR conversation for in-progress screenshots gathered during development; a fresh before/after pair for the final state has not been attached to this description
- [ ] If version bumped: all three files updated together - not applicable, no version bump in this PR
- [x] Relevant documentation updated to reflect changes (`DESIGN.md` added, `docs/insights-backend-brief.md` added, `AGENTS.md`/`CLAUDE.md` updated)
- [x] Risks documented above

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,1 +1,1 @@
-See [CLAUDE.md](./CLAUDE.md) for all project context, conventions, and instructions.
+See [CLAUDE.md](./CLAUDE.md) for all project context, conventions, and instructions. Before adding or changing frontend controls, also read [DESIGN.md](./DESIGN.md): it defines the global button and dropdown primitives, animation rules, and control inventory.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -28,7 +28,9 @@ Verenu is an open-source AI dictation desktop app for Windows and macOS — a fr
 
 ## Design System
 
-See [docs/colors.md](docs/colors.md) for the complete color palette, typography, and theming system.
+Read [DESIGN.md](DESIGN.md) before creating or restyling frontend controls. It is the source of truth for shared button variants, dropdown anatomy, UI motion, and the complete control inventory. See [docs/colors.md](docs/colors.md) for the complete color palette, typography, and theming system.
+
+Shared interactive styles live in `src/ui.css`; do not recreate `.btn-primary`, `.btn-ghost`, `.btn-danger`, or `.ui-dropdown-*` styles inside a Svelte component. Preserve feature-specific controls when they communicate a distinct state, such as navigation, recording, selection cards, or system permissions.
 
 ## Build & Dev Commands
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,0 +1,127 @@
+# Verenu UI Design System
+
+This is the source of truth for shared frontend controls. Use it before adding or changing a button, dropdown, toggle, or control animation.
+
+## Source of truth
+
+| File | Owns |
+| --- | --- |
+| [`src/theme.css`](src/theme.css) | Themeable color, typography, radius, layout, and base elevation tokens. |
+| [`src/ui.css`](src/ui.css) | Shared button states, dropdown anatomy, focus feedback, and CSS motion tokens. |
+| [`src/lib/motion.ts`](src/lib/motion.ts) | Svelte transition timing, distance, reduced-motion behavior, and reusable transition helpers. |
+| [`src/lib/components/Dropdown.svelte`](src/lib/components/Dropdown.svelte) | Headless outside-click and Escape handling for composed dropdowns. |
+| [`src/lib/components/Toggle.svelte`](src/lib/components/Toggle.svelte) | The shared accessible switch control. The `toggle` class is a smoke-test contract. |
+| [`docs/colors.md`](docs/colors.md) | Palette and typography reference. |
+
+Do not define `.btn-primary`, `.btn-ghost`, `.btn-danger`, `.ui-dropdown-*`, or their basic interactive states inside a Svelte component. Add a semantic modifier only for layout or content that the primitive cannot know, such as a menu width or text truncation.
+
+## Standard controls
+
+### Buttons
+
+```svelte
+<!-- One obvious next action. -->
+<button class="btn-primary" onclick={save}>Save changes</button>
+
+<!-- A secondary, non-destructive action. -->
+<button class="btn-ghost" onclick={cancel}>Cancel</button>
+
+<!-- A destructive action. Use an explicit label. -->
+<button class="btn-danger" onclick={deleteItem}>Delete history</button>
+
+<!-- Use only when the surrounding control is intentionally compact. -->
+<button class="btn-primary btn-compact" onclick={add}>Add</button>
+```
+
+| Variant | Use for | Shared behavior |
+| --- | --- | --- |
+| `btn-primary` | The main action in a local context, such as Save, Add, or Confirm. | Ink surface, strong contrast, hover, disabled state, focus ring. |
+| `btn-ghost` | Secondary actions, utilities, external links, and dismiss actions. | Transparent surface, bordered hover state, disabled state, focus ring. |
+| `btn-danger` | Confirmed destructive actions only. | Warning surface, filled destructive hover, disabled state, focus ring. |
+| `btn-compact` | A space-constrained button that still has a complete label. | Reduces height, padding, and type size. It is a size modifier, not a visual variant. |
+
+Do not use a standard button for navigation rows, selection cards, tabs, microphone recording, or the dictation pill. Those controls communicate different states and are intentionally specialized.
+
+### Dropdowns
+
+Use the structural classes on a feature-local dropdown. This centralizes visual treatment while allowing each feature to retain typed values, option descriptions, width constraints, and keyboard behavior.
+
+```svelte
+<script lang="ts">
+  import Dropdown from '$lib/components/Dropdown.svelte';
+
+  let open = $state(false);
+</script>
+
+<Dropdown bind:open closeSelector=".example-dropdown">
+  <div class="ui-dropdown example-dropdown">
+    <button
+      class="btn-ghost ui-dropdown-trigger"
+      aria-haspopup="listbox"
+      aria-expanded={open}
+      onclick={() => (open = !open)}
+    >
+      Current option
+      <svg aria-hidden="true">...</svg>
+    </button>
+
+    {#if open}
+      <div class="ui-dropdown-menu" role="listbox">
+        <button class="ui-dropdown-option active" role="option" aria-selected="true">Current option</button>
+      </div>
+    {/if}
+  </div>
+</Dropdown>
+```
+
+| Class | Responsibility |
+| --- | --- |
+| `ui-dropdown` | Anchors a menu to its trigger. |
+| `ui-dropdown-trigger` | Trigger shape, chevron motion, focus ring, and open treatment. Combine with `btn-ghost` when it is a secondary setting control. |
+| `ui-dropdown-menu` | Popover surface, stacking, border, shadow, and scroll behavior. |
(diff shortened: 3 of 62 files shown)


MONKE2525E:
Addressed all 18 review findings in \

MONKE2525E:
Round 2 addressed in \

MONKE2525E:
Round 3 addressed in ee3219a9. All 17 actionable findings fixed (1 kept):

**Backend**
1. session.rs — when a new session is active, the cancelled-capture pill is skipped without touching the live pill; hide_pill now only fires on lock-poisoning (this also fixed a regression my round-2 fix introduced)
2. injection/windows.rs — explicit target_thread_id != 0 guard (finding 2 from round 2, now verified)
3. injection/windows.rs — the UIA probe is now always fetched, so the container-focus guard runs even when contextual_caps/auto_spacing are off
4. insights.rs — STOPWORDS lookup via binary_search (sorted, guarded by a sortedness test); /transcription suffix stripped in the primary= branch too
5. transcriptions.rs — lifetime_stats upserts so a fresh DB with no id=1 row still counts words/dictionary fixes

**Frontend**
6. Insights.svelte — listener torn down on unmount even if it resolves late; no state mutation after unmount; transcribed refresh is silent
7. DonutChart.svelte — canvas resize now precedes the DPR transform (fixes my round-2 Retina regression); destroyed probe nulled out
8. StreakHeatmap.svelte — lastMonth advances on every boundary (labels never land mid-month); bestWeekday counts only active days
9. DailyChart/StreakHeatmap — reduce instead of Math.max(...) for long series
10. Dropdown.svelte — instanceof Element guard before .closest()
11. CostBreakdown — usage table stays visible when all models are unpriced (donut hidden only)
12. HeroStats — pluralize books off the displayed amount
13. helpers.ts — fmtCompact thresholds on rounded values (999.6 to 1.0k); fmtUsd sign placement and -0.00 to \.00
14. pricing.ts — task normalized before override lookup

**Kept as-is (1)** — HWND(target_hwnd as *mut core::ffi::c_void) cast: it matches the existing codebase pattern, compiles and runs on Windows CI; the suggestion was stylistic.

Checks: svelte-check clean, cargo clippy with -D warnings clean, full test suite 480 passed.

MONKE2525E:
Round 4 addressed in b286c9d1. All 6 findings fixed:

1. \hotkey/win.rs\ — Ctrl+Alt+C is only swallowed when a copy callback is actually registered; with none set the keystroke falls through to the target app
2. \insights.rs\ — a quiet today no longer zeroes a streak that's alive through yesterday (GitHub-style grace), with two new tests covering the grace and the two-quiet-day break
3. \injection/windows.rs\ — removed the categorical pane/window container abort. You're right: Qt, Flutter, Swing, terminals, and custom Win32 controls expose only a generic pane/window to UIA with UnsupportedControl yet accept pastes, so the abort broke them. The \GetGUIThreadInfo\ \has_focus\ guard already rejects the nothing-focused case; the probe now only feeds contextual formatting and post-paste verification.
4. \session.rs\ — a quality-gate rejection no longer calls \hide_pill\ if a new session already started while stopping
5. \PillApp.svelte\ — the \copied\ state keeps its \errorMsg\ (carried via pill-error just before pill-state) instead of clearing it
6. \Dropdown.svelte\ — a Text-node click target is resolved to its parent element before the \closest(closeSelector)\ check

Checks: svelte-check clean, cargo clippy with -D warnings clean, full suite 482 passed.

MONKE2525E:
Round 5 addressed in 71dbc9b8. All 5 findings fixed:

1. \insights.rs\ — the quiet-today streak grace was buggy: when today was active, the pending flag stayed set and could swallow real historical gaps. The grace now applies only to the trailing day itself; added a regression test (today active + yesterday gap → streak of 1)
2. \pp_hotkey.rs\ — CopyLast hotkey uses \	ry_state::<DbHandle>()\ and logs+skips if state is gone (a hook can fire during teardown)
3. \
ecording.rs\ — the detached \cancel_recording_with_resume\ task's join handle is awaited in a side task so panics/aborts are logged instead of silently swallowed
4. \Insights.svelte\ — range-picker clicks now stopPropagation, matching the notification-dropdown pattern, so the trigger toggle no longer races the window click-outside handler
5. \DonutChart.svelte\ — the tiny-sweep inset is bounded to sweep/3 so a sub-0.02 radian segment can't invert into a full-circle arc, and the exact-annulus sector is used as a fallback when the fillet path can't be built

Checks: svelte-check clean, cargo clippy with -D warnings clean, full suite 483 passed.

MONKE2525E:
Round 6 addressed in 7383fb2f. All 7 findings fixed:

1. \hotkey/win.rs\ — the C keyup is only swallowed while the Ctrl/Alt chord is still held; if the modifiers were released first, the keyup passes through so the target app's key state stays consistent
2. \	auri.ts\ — the dev-insights fixture tracks longestStartedOn/longestEndedOn while scanning the run, instead of brittle index arithmetic
3. \HistoryList.svelte\ — the discard button is disabled while \
esumingCancelled\ so it can't race the resume
4. \HistoryList.svelte\ — the first row after the banner gets its \irst-in-table\ top border (the Today header is omitted from the virtual list when the banner shows)
5. \DonutChart.svelte\ — a stale \hoveredId\ is cleared when its segment is removed, so a data change no longer locks the chart/legend in a dimmed state
6. \HeroStats.svelte\ — pluralization keys off the formatted \ooksLabel\ value, so 80,080 words (books = 1.001) reads '1 full-length book', not '1.0 books'
7. \HourStrip.svelte\ — a missing/non-array \hourly\ prop falls back to a 24-slot zero array

Checks: svelte-check clean, cargo clippy with -D warnings clean, full suite 483 passed.

MONKE2525E:
Round 7 addressed in e589a74d. All 3 findings fixed:

1. \Insights.svelte\ — a silent refresh that fails while \data\ is still null now surfaces the error state instead of leaving the skeleton loading forever
2. \HistoryList.svelte\ — restored the \!hasBanner\ guard on the first row's \irst-in-table\ border; the banner's day-table already draws the separator, so \index === 0\ alone produced a double border
3. \DonutChart.svelte\ — removed the redundant \startValueTween()\ from onMount; the \segments\ effect already runs it on mount, and the second call reset \nimFrom\ to the target and skipped the entrance animation

Checks: svelte-check clean, cargo clippy with -D warnings clean.

MONKE2525E:
Round 8 addressed in e4675b21. All 7 findings fixed:

1. \
ecording.rs\/\state.rs\ — the paste-failure text is now peeked (not taken) and only cleared after the clipboard write succeeds, so a transient clipboard error keeps the retry alive
2. \PillApp.svelte\ — the pill stays up while \
etry_transcription\ is in flight; it only goes idle if the retry itself fails, matching the cancelled-resume pattern
3. \Insights.svelte\ — a silent refresh failure now surfaces the error when there's no data, and restores \status = 'loaded'\ when good data exists so a superseded non-silent load can't leave the skeleton stuck
4. \DonutChart.svelte\ — the color probe's inline color is cleared before assigning, so an invalid CSS color can't leak a previous resolve
5. \HeroStats.svelte\ — the wpm gauge value renders only when > 0 (explicit), not via falsiness
6. \helpers.ts\ — \pctDelta\ guards against NaN/undefined previous values
7. \helpers.ts\ — \parseLocalDay\ rejects malformed/empty dates (including out-of-range calendar dates) with an Invalid Date so \mtDay\/\mtDayLong\ fall back to the raw string instead of a silent year-1900

Checks: svelte-check clean, cargo clippy with -D warnings clean, full suite 485 passed.

MONKE2525E:
Round 9 addressed in 400665ac. All actionable findings fixed (1 was a self-corrected false positive):

1. \
ecording.rs\/\state.rs\ — reserve the start before consuming the cancelled capture, attach the peeked audio as prepend, and only clear the stashed slot once the session is actually underway. A mid-start failure (mic permissions, poisoned lock) now leaves the audio resumable.
2. \win.rs\ — acknowledged as a false positive: \modifier_held(91)\ already checks both 91 and 92, so no change needed
3. \insights.rs\ — \
ormalized\ is moved into \counts.entry\ instead of cloned
4. \PillApp.svelte\ — a stale \copiedTimer\ is cleared on paste_failed entry/exit so it can't fire \goIdle()\ on a fresh state
5. \PillApp.svelte\ — a failed retry only goes idle while still in the \error\ state (the pill may have moved to an active state during the invoke)
6. \helpers.ts\ — \mtDuration\ and \
iceCeiling\ guard non-finite/negative inputs

Checks: svelte-check clean, cargo check clean.

PR #211, opened 2026-08-12
Polish dictionary and snippets toolbar search field

MONKE2525E:
The dictionary and snippets toolbar search field rendered as a bordered box floating inside its elevated strip, and the sort pills and clear button had no keyboard focus treatment.

What changed:

- Swap the hand-rolled search input for the shared `ui-input` dense variant and remove the 1px border around the field so it blends into the strip. Keyboard focus still shows the accent border and ring via an explicit `:focus-visible` rule.
- Add the shared focus ring to the sort pills and clear button.
- Use the `--r-sm` radius token instead of hardcoded 8px/5px values.

Verified: `svelte-check` clean; dictionary smoke test 22/22 passing (frozen contract unchanged â€” `sort-pill`, `search`, `btn-primary` classes kept); live browser check on both the Dictionary and Snippets pages shows the borderless field with the same 44/32/30 control heights as before.

### DIFF
--- a/src/lib/views/dictionary/DictionaryToolbar.svelte
+++ b/src/lib/views/dictionary/DictionaryToolbar.svelte
@@ -52,14 +52,14 @@
       <circle cx="11" cy="11" r="7"/><path d="m21 21-4.35-4.35"/>
     </svg>
     <input
-      class="search-input"
+      class="ui-input ui-input--dense"
       type="text"
       placeholder={`Search ${count} ${count === 1 ? 'term' : 'terms'}…`}
       bind:value={search}
       aria-label="Search dictionary"
     />
     {#if search}
-      <button class="clear-btn" onclick={() => search = ''} aria-label="Clear">
+      <button class="clear-btn ui-focus-ring" onclick={() => search = ''} aria-label="Clear">
         <svg width="10" height="10" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.5" stroke-linecap="round"><path d="M18 6 6 18M6 6l12 12"/></svg>
       </button>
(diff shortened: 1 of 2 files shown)


MONKE2525E (commented):
Self-review (AI review job is currently broken — CLIPROXY_MODEL is empty in the workflow env, so the provider call fails before it reads the diff).

Verdict: no blocking findings. Change is small, focused, and consistent across both toolbars.

Verified:
- Focus ring preserved: the scoped .search .ui-input override (0,3,0) out-specifies ui.css's :where(...).ui-input:focus-visible (0,2,0), so the explicit .search .ui-input:focus-visible companion rule is required — present in both files, matching the shared rule exactly (accent border + --ui-focus-ring).
- No layout shift: transparent border keeps the input at the same 32px height (border-box), strip stays 44px, pills 32px, button 30px.
- Parity: both toolbar files got identical CSS changes; only pre-existing structural differences remain (props, sort keys, labels).
- Contracts: no test references the removed .search-input class (lang-search-input / app-search-input are separate components); .sort-pill/.search/.clear-btn/.btn-primary preserved. Local smoke suites pass; CI All-in-one fast profile passed (4m45s).
- Design system: radii now tokenized (--r-sm and calc(--r-sm - 3px) compute to the same 8px/5px); input uses the shared ui-input dense primitive; focus treatment follows ui.css conventions.

Nits (pre-existing, out of scope):
- The two toolbars still duplicate ~90 lines of identical CSS; per DESIGN.md adoption rule 3, extract only after a third use.
- .search .ui-input background: transparent duplicates the dense variant's background inside the strip — harmless, strip is --bg-elev in both themes.

PR #224, opened 2026-08-15
Add legacy pages toggle and context website checks

MONKE2525E:
## Thinking Path

> - Verenu is a Windows and macOS AI dictation app.
> - This PR covers the settings, contexts, legacy-page navigation, and related documentation subsystems.
> - The legacy-page toggle was promised but unavailable, context inputs lacked a client-side name limit, and website targets were not validated.
> - The changes are needed to make the new Contexts workflow usable while preserving access to the older pages.
> - This pull request adds the Legacy mode toggle, context validation and feedback, targeted animations, and DNS-only website checks.
> - The benefit is clearer navigation, safer context setup, and documented local-first privacy behavior.

## UI Screenshots

Before — Contexts is the primary vocabulary/snippets workflow:

Before: Contexts navigation

After — enabling Legacy pages restores Dictionary and Snippets in the primary navigation:

After: Legacy navigation

## What Changed

- Add a persisted Legacy pages setting with confirmation UI, navigation gating, and redirects between Contexts and legacy pages.
- Add Contexts documentation and update legacy feature/privacy/troubleshooting references.
- Enforce a 30-character context-name limit with a live counter.
- Fix App Mappings entrance animation stutter and color-picker sizing.
- Validate website targets with a bounded DNS-only lookup and update dev/test mocks.
- Animate newly added app and website chips without replaying animations for existing selections.

## Verification

- `npm run check` — passes with 0 errors and 0 warnings.
- `cargo check` in `src-tauri/` — passes.
- Existing smoke coverage was updated for Legacy mode and the summary reports the suite passing except the pre-existing About-panel repository-link mismatch.

## Risks

- Low-to-moderate UI risk: navigation changes are gated by a persisted opt-in setting and redirect invalid pages.
- Website checks are DNS-only, bounded to four seconds, and fail closed for an unresolved domain while allowing actual lookup errors through.
- No API keys, dictation data, or history data are sent by the website check.

## Model Used

- OpenAI GPT-5.6 Luna via Codex, tool-assisted coding mode.

## Checklist

- [x] Thinking path traces from Verenu context down to this specific change
- [x] Model used is specified (provider + version)
- [x] Relevant documentation updated to reflect changes
- [x] `npm run check` passes (TypeScript)
- [x] No API keys, secrets, or personal data in code or logs
- [x] Risks documented above

PR #231, opened 2026-08-15
Remove Gemini 3.5 Flash from saved model lists

MONKE2525E:
## Summary\n\nReplace persisted Gemini 3.5 Flash entries with Gemini 3.7 Flash so the obsolete model no longer appears in the model picker. Update Gemini recommendations and migrate defaults and fallbacks.\n\n## Verification\n\n- npm run check\n- Focused Rust API tests passed previously\n

---
This is **part 2 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #224
- <kbd>&nbsp;2&nbsp;</kbd> #231 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #233

### DIFF
--- a/src/lib/components/settings/ModelsSection.svelte
+++ b/src/lib/components/settings/ModelsSection.svelte
@@ -52,6 +52,7 @@
     modelId,
     recommendedModels,
     migrateDeprecatedGroqCleanupModel,
+    migrateDeprecatedGoogleModel,
     splitModelId,
     type AllSettingsPayload,
     type TaskType,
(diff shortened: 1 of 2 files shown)


MONKE2525E:
## Blocking finding\n\nModelsSection.svelte migrates provider-prefixed Google IDs for transcription defaults and transcription fallbacks, but not for cleanup defaults or cleanup fallbacks.\n\nAt lines 568-576, cleanupDefaultModel is passed through migrateDeprecatedGoogleModel before parsing, so google/gemini-3.5-flash does not match the bare-ID helper and then only the Groq migration runs. Lines 590 onward similarly migrate cleanup fallbacks only for Groq. A saved cleanup default or fallback of google/gemini-3.5-flash can therefore be re-added by ensureDefaultAndFallbacks() and still appear in the picker. Please apply the same Google provider migration to both cleanup paths, ideally through one shared provider-model normalization helper.

MONKE2525E:
Fixed in commit 2da9ad0: cleanup defaults and cleanup fallbacks now migrate provider-prefixed google/gemini-3.5-flash IDs to google/gemini-3.7-flash. 
pm run check passes with 0 errors and warnings. Manual re-review found no remaining code issues.

PR #372, opened 2026-09-08
Fix audit regressions and bound resource use

MONKE2525E:
## Thinking Path

> - Verenu is a Windows and macOS AI dictation app: hotkey hold -> audio capture -> transcription -> cleanup -> text injection.
> - This PR focuses on the data, recovery, auto-learn, and native-build paths that retain work or do repeated expensive work.
> - The reviewed fixes covered unbounded history/FTS state, stale sync deletes, repeated Insights scans, unsafe text anchoring, recovery persistence, and Windows test startup.
> - These paths can silently lose corrections, resurrect deleted data, pause the app, or fail before tests start.
> - This pull request makes the operations bounded or transactional and adds regression coverage around the failure cases.
> - The benefit is predictable memory/storage use, safer recovery and sync behavior, and reliable Windows verification.

## What Changed

- Repaired FTS initialization and recovery atomically, including trigger/readiness validation and direct MATCH queries.
- Consolidated Insights text metrics into one streaming database scan while preserving existing normalization rules.
- Made auto-learn range reads bounded and resilient to caret movement, long injections, shifted windows, and invalid anchors.
- Preserved durable sync tombstones while allowing stale peers to resnapshot without resurrecting deleted rows.
- Hardened persistence and recording recovery behavior under queue pressure and disk failures.
- Added regression coverage for history pagination, FTS repair, Insights metrics, auto-learn anchoring, sync deletion handling, failover, and recovery paths.
- Added a shared Windows activation manifest for app and Rust test executables without duplicate manifest resources.

## Verification

- `npm run lint` passed.
- `npm run test:unit` passed: 17 files, 134 tests.
- `cargo test --quiet` passed: 734 tests, 4 intentionally ignored.
- Windows binary and library test executables both linked and ran successfully.
- Embedded Common Controls v6 manifests were inspected in both Windows executables.

## Risks

- Sync tombstones remain durable per row so old or re-paired devices cannot resurrect deletions; this bounds log history while retaining the latest row state.
- Live microphone hardware behavior and macOS runtime behavior were not exercised in this Windows session.
- Existing unrelated worktree changes were excluded from this PR.

## Model Used

- OpenAI Codex, gpt-5.6-luna, tool use and extended repository verification.

## Checklist

- [x] This PR targets `master` directly
- [x] Model used is specified
- [x] `npm run check` passes
- [x] `npm run lint` passes
- [x] `npm run test:rust` equivalent full `cargo test --quiet` passes
- [ ] Smoke tests pass (not run; native app instance was left running)
- [x] Risks documented above


MONKE2525E:
I reviewed every finding against the current PR head `540a54dbc` rather than treating the `info` label as severity.

The two findings that were still actionable are fixed in this head:

- `src-tauri/src/api/auto_learn/correction.rs`: `current_span_from_surrounding_context` now validates ordering, UTF-8 boundaries, and range validity before slicing. Malformed or stale anchors return `None` instead of panicking.
- `src-tauri/src/data/db/insights.rs`: top-word ranking now collects and sorts once, then truncates to 12 entries, avoiding a sort and pop on every distinct word.

The other five findings are already fixed in the current code:

- `build.rs`: `link_windows_manifest` and its call are no longer host-`cfg` gated, so target-aware cross-compilation handling runs.
- `data/db/snippets.rs`: insert, update, delete, and context-move mutations invalidate the snippet cache.
- `pipeline/failover.rs`: loaded recovery metadata is recalculated from the samples actually read.
- `pipeline/mod.rs`: the debug size log derives WAV size from the captured PCM sample count instead of the lazy, not-yet-materialized buffer.
- `sync/engine.rs`: tombstone progress uses the selected row sequence directly and no longer performs the redundant row lookup.

Please run a fresh review against the current head.

